PR #3, opened 2020-03-25
Relicensing and Renaming

ProfFan:


dellaert:
Can you split the PR? I'd like to understand the Jacobian thing and separate it from name change and license...

ProfFan:
Yea I can do that, just one minute :)

ProfFan:
@dellaert Ping

PR #4, opened 2020-03-25
Simplified interface for Double jacobians

ProfFan:
This PR added simple inference for `basisVectors` so you can just write `jacobian(of: at:)` instead of `jacobian(of: at: basisVectors:)`.


ProfFan:
@dellaert Here it is

Note that these tests are not real tests-they print out the jacobian so we can make design choices :P

dellaert:
"No description provided". Please provide one - I don't understand what you are trying to do. We probably need to discuss in person.

PR #5, opened 2020-03-25
Add initial CI

ProfFan:


dellaert:
I should not see force-push on public branches I think :-)

PR #14, opened 2020-03-31
Unified jacobian protocol

### LINKED ISSUE #13: Unified Jacobian Protocol that can handle both complex and primitive types
I always find this quite annoying when implementing generic stuff related to `TangentVector`:

In JacobianProtocol you will find out that I am using different generics for `Double` and non-`Double` types, and I tried to make it more generic by introducing the `JacobianEvaluatable` protocol. However, the `basisVectors()` method only works for tangent types that is complex, so the current implementation will fail for tangents of `[Rot2]`, which is `[Double]`.

marcrasi:
This adds a `TangentStandardBasis` protocol that defines a basis for the tangent space of a type. This protocol allows us to unify all the `jacobian` functions into one function that works for all types.

It would be simpler to have a `StandardBasis` protocol that defines a basis for a vector space, and then conform `TangentVector` to `StandardBasis`. But we can't do that right now because of https://bugs.swift.org/browse/TF-1227. I think we should do that after https://bugs.swift.org/browse/TF-1227 gets fixed.

This resolves #13.

### DIFF
--- a/Sources/SwiftFusion/Core/Jacobian.swift
+++ b/Sources/SwiftFusion/Core/Jacobian.swift
@@ -1,7 +1,3 @@
-public protocol JacobianEvaluatable: Differentiable, KeyPathIterable where TangentVector: KeyPathIterable {
-  
-}
-
 // A jacobian can be calculated by applying the pullback (reverse mode) to the basis vectors of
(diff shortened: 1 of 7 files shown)


ProfFan:
Thanks!

PR #16, opened 2020-04-01
change Pose2 derivative convention to match GTSAM

### LINKED ISSUE #8: Derivative is not consistent with GTSAM
I have added a test to compare the Jacobian generated by the autodiff system and GTSAM, and they do not match. I am still investigating why this is happening.

Sample:

For Pose2, expected is a 3x2 matrix, values in Pose2's tangent space

In GTSAM, it corresponds to
```
Matrix expectedH1 = (Matrix(3,3) <<
    0.0,-1.0,-2.0,
    1.0, 0.0,-2.0,
    0.0, 0.0,-1.0
).finished();
```
In current SwiftFusion, it is
```
[0.0, -1.0,  2.0]
[1.0,  0.0, -2.0]
[0.0,  0.0, -1.0]
```
```
Matrix expectedH2 = (Matrix(3,3) <<
     1.0, 0.0, 0.0,
     0.0, 1.0, 0.0,
     0.0, 0.0, 1.0
).finished();
```
In current SwiftFusion, it is
```
 0.0, 1.0, 0.0
-1.0, 0.0, 0.0
 0.0, 0.0, 1.0
```

marcrasi:
Defines custom derivatives for `Pose2.init`, `Pose2.rot`, and `Pose2.t`, following the convention from [1] section 5.1.

Adds some tests verifying that derivatives match some of the the theorems about derivatives in [1]: derivatives of left and right multiplication from sections 5.2 and 5.4, and derivatives of the inverse from section 5.3.

Resolves #8.

[1] https://github.com/borglab/gtsam/blob/develop/doc/math.pdf


Sources/SwiftFusion/Geometry/Rot2.swift
```diff
@@ -135,12 +135,20 @@ public extension Rot2 {
     Point2(c * p.x + -s * p.y, s * p.x + c * p.y)
   }
 
+  func rotate(_ p: Point2.TangentVector) -> Point2.TangentVector {
```
dellaert, line 138: Hmm, SO(2) typically acts on Point2, so is this a *second* rotate for Point2.TangentVector? Of course, they are the same, because Point2 is a vector space. in fact, in GTSAM we had a special Point2 type at first but we deprecated it in favor of just equating it to Vector2. In which case we do not need a “special” rotate for TangentVectors. (see VectorSpace.h in GTSAM)

Tests/SwiftFusionTests/Core/JacobianTests.swift
```diff
@@ -39,21 +40,16 @@ class JacobianTests: XCTestCase {
...
-    XCTAssert(
-      expected.recursivelyAllKeyPaths(to:Double.self)
-        .map {j[keyPath: $0] - expected[keyPath: $0]}
-        .reduce(0.0, {$0 + abs($1)}) < 1e-10
-    )
-  }
-  
+
+    let expected = Tensor<Double>([
+      [0.0, -1.0, -2.0, 1.0, 0.0, 0.0],
```
dellaert, line 45: Should this stuff not be in Pose2Tests? Also, we should probably discuss the return type. This is a two argument function, so in GTSAM we would compute 2 3x3 Jacobians. I imagine we will refine/change this as we do a TDD development of Gauss-Newton optimization. For CGLS we don’t need this probably.

Tests/SwiftFusionTests/Geometry/Pose2Tests.swift
```diff
@@ -1,6 +1,24 @@
-import SwiftFusion
+import TensorFlow
 import XCTest
 
+import SwiftFusion
+
+func randomPose2() -> Pose2 {
```
dellaert, line 6: Make a method Random in Pose2? Arbitrary tangentvector with given covariance matrix and then move :-)

Tests/SwiftFusionTests/Geometry/Pose2Tests.swift
```diff
@@ -1,6 +1,24 @@
...
+
+func eye3x3() -> Tensor<Double> {
+  Tensor<Double>([
+    [1, 0, 0],
+    [0, 1, 0],
+    [0, 0, 1]
+  ])
+}
+
+func adjointMatrix(_ x: Pose2) -> Tensor<Double> {
```
dellaert, line 18: Method in Pose2?

Tests/SwiftFusionTests/Geometry/Pose2Tests.swift
```diff
@@ -119,15 +137,60 @@ final class Pose2Tests: XCTestCase {
     let p5T1 = between(map[4], map[0])
 
     // Test condition: P_5 should be identical to P_1 (close loop)
-    XCTAssertEqual(p5T1.t_.magnitude, 0.0, accuracy: 1e-2)
+    XCTAssertEqual(p5T1.t.magnitude, 0.0, accuracy: 1e-2)
+  }
+
+  func testDerivativeIdentity() {
```
dellaert, line 143: Document

Tests/SwiftFusionTests/Geometry/Pose2Tests.swift
```diff
@@ -181,14 +244,17 @@ final class Pose2Tests: XCTestCase {
...
-    XCTAssertEqual(p5T1.t_.magnitude, 0.0, accuracy: 1e-2)
+    XCTAssertEqual(p5T1.t.magnitude, 0.0, accuracy: 1e-2)
   }
 
   static var allTests = [
     ("testBetweenIdentitiesTrivial", testBetweenIdentitiesTrivial),
     ("testBetweenIdentities", testBetweenIdentities),
     ("testBetweenIdentities", testBetweenIdentitiesRotated),
     ("testBetweenDerivatives", testBetweenDerivatives),
+    ("testDerivativeIdentity", testDerivativeIdentity),
```
dellaert, line 255: That’s not automatic?

Tests/SwiftFusionTests/TestUtilities.swift
```diff
@@ -0,0 +1,29 @@
+import TensorFlow
```
dellaert, line 1: Document everything ;-)

Sources/SwiftFusion/Geometry/Pose2.swift
```diff
@@ -1,13 +1,32 @@
...
+  private var rot_: Rot2
+
+  @differentiable public var t: Point2 { t_ }
+  @differentiable public var rot: Rot2 { rot_ }
+
+  public struct TangentVector:
+    Differentiable, AdditiveArithmetic, KeyPathIterable, VectorProtocol, ElementaryFunctions
+  {
+    public typealias VectorSpaceScalar = Double
+    public var t: Point2.TangentVector
```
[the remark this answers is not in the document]
marcrasi: I see I have sort of unthinkingly used the `Point2.TangentVector` and `Rot2.TangentVector` types here to store the components of the tangent space. This doesn't really make sense, as you say. So I'll change this struct definition to define its own components. We don't get much from using `Point2.TangentVector` and `Rot2.TangentVector`.

Sources/SwiftFusion/Geometry/Rot2.swift
marcrasi: I won't need these after I stop using `Point2.TangentVector` inside of `Pose2.TangentVector`, so I'll remove them!

Sources/SwiftFusion/Geometry/Pose2.swift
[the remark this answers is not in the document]
dellaert: Actually, there are two concepts related to manifolds: the “true tangent space” which in this case is a set of special 3x3 matrices, and the “vector tangent space” that is isometric to it (via the `hat`  and `vee`  maps. In GTSAM we just use the latter, and so really the *only* thing we have to specify here - I think- is the dimensionality of the tangent space. Since this is *always* the same as the dimensionality of the manifold, even that should be derived from the Manifold itself. Less code = good ? :-)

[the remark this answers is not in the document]
dellaert: *isomorphic

Sources/SwiftFusion/Geometry/Rot2.swift
dellaert: Still, I think my comments holds. I think we should simply identify Point2 with Vector2. Of course, that assumes we have a generic Vector class :-)

marcrasi: Yes, I'll file a ticket for a vector type :)

dellaert: Can we just use 1D tensors?

Sources/SwiftFusion/Geometry/Pose2.swift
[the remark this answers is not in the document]
marcrasi: Yes, I can sort of imagine one day being able to specify a few properties of a Lie group (dimension, exp, special names for certain algebra basis vectors, ...?) and have a library (possibly with compiler support) synthesize all the required types and derivatives automatically.

Ideally this would be mostly just a library, maybe with a few hooks in the compiler so that the library can control the synthesis more than is possible today. (Right now Swift will synthesize everything you need when you manifold is a direct product of manifolds, but it can't do anything else.)

I will start thinking about how something like this would be possible. I think it'll be a while before I come up with anything because I'm still very new to Lie groups and I'll have to keep playing with them and reading up on how they work in GTSAM before I figure anything out.

In the meantime, I'd like to just explicitly declare the type of the tangent space.

Tests/SwiftFusionTests/Core/JacobianTests.swift
marcrasi: A `jacobian` that deals with 2-argument functions by returning 2 jacobian matrices should be pretty easy to define! I'll file a ticket for this.

marcrasi: Also I'll move these to `Pose2Tests`

Sources/SwiftFusion/Geometry/Rot2.swift
marcrasi: Do you mean plain unwrapped `Tensor` from `TensorFlow`?

That would work, but it has no static shape information, which would make the geometry apis more confusing/errorprone because it's harder to see what size vector you have and what size vector you need.

So I would prefer having something like Vector2, Vector3, ..., Vectorn, and Vector. (Unfortunately Swift doesn't have any way of being generic over integer values right now.) These could be wrappers around `Tensor`.

dellaert: Sure. FYI in GTSAM `Vector3` is a typedef to a fixed-size Eigen Vector. Having Vector2 and Vector3 here would be very familiar :-)

Sources/SwiftFusion/Geometry/Pose2.swift
```diff
@@ -16,19 +35,51 @@ public struct Pose2: Equatable, Differentiable, KeyPathIterable, TangentStandard
...
+  static func vjpInit(_ r: Rot2, _ t: Point2)
+    -> (value: Pose2, pullback: (TangentVector) -> (Rot2.TangentVector, Point2.TangentVector))
+  {
+    return (Pose2(r, t), { ($0.rot, r.rotate($0.t)) })
+  }
+}
+
+extension Pose2 {
+  /// The adjoint representation of `self`.
+  public func adjoint(_ v: TangentVector) -> TangentVector {
```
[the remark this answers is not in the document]
marcrasi: Are the two different ones the "adjoint map" and the "adjoint representation"? Those are the only two adjoints I've found so far, but I'm not very familiar with this stuff yet.

If it's that we could name this `adjointRepresentation` to disambiguate it.

Sources/SwiftFusion/Geometry/Pose2.swift
```diff
@@ -16,19 +35,51 @@ public struct Pose2: Equatable, Differentiable, KeyPathIterable, TangentStandard
...
+    return (t_, { TangentVector(t: rot_.unrotate($0), rot: .zero) })
+  }
+
+  @derivative(of: rot)
+  @usableFromInline
+  func vjpRot() -> (value: Rot2, pullback: (Rot2.TangentVector) -> TangentVector) {
+    return (rot_, { TangentVector(t: .zero, rot: $0) })
+  }
+
+  @derivative(of: init(_:_:))
```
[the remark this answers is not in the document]
marcrasi: I've added tons of comments explaining what's going on :)

Tests/SwiftFusionTests/Geometry/Pose2Tests.swift
marcrasi: Nope, unfortunately swift can't automatically detect the tests on Linux for some reason. I think I heard someone is working on it, so hopefully this will get improved soon.

Sources/SwiftFusion/Geometry/Pose2.swift
```diff
@@ -16,19 +35,51 @@ public struct Pose2: Equatable, Differentiable, KeyPathIterable, TangentStandard
     rot_ = r
   }
 
+  @derivative(of: t)
```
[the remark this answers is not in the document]
marcrasi: added :)

Tests/SwiftFusionTests/Geometry/Pose2Tests.swift
marcrasi: done

Tests/SwiftFusionTests/TestUtilities.swift
marcrasi: done

marcrasi:
Thanks for all the comments! This is pretty fun.

I've addressed all your comments except for the `adjoint` naming question. I asked you about that inline.

I discovered that it will be very nice to have a `Vector2` type because there are a lot of places in this PR where I'm doing rotations of `Vector2`-like things and I have to type out the rotation transformation every time. But I would like to add `Vector2` later in a separate PR, to keep this one reasonably small.

Tests/SwiftFusionTests/Geometry/Pose2Tests.swift
marcrasi: done

Tests/SwiftFusionTests/Geometry/Pose2Tests.swift
marcrasi: done

Sources/SwiftFusion/Geometry/Pose2.swift
```diff
@@ -1,34 +1,223 @@
...
+  {
+    // Explanation of this calculation:
+    //
+    // We would like the differential of `init` at `(t, r)`, `dinit_(t, r)`, to satisfy:
+    //   init(t * exp(eps_t), r * exp(eps_r)) ~= init(t, r) * exp(dinit_(t, r)(eps_t, eps_r))
+    // where `eps_t`, `eps_r` are "small" elements of R^2 and so(2) respectively.
+    //
+    // Multiplying on the left by `init(t, r)^-1` gives
+    //   exp(dinit_(t, r)(eps_t, eps_r))
+    //     = init(t, r)^-1 * init(t * exp(eps_t), r * exp(eps_r))
```
dellaert, line 64: I'd state `t * exp(eps_t) == t + eps_t` and `inverse(r) = r^t` and use that
Also, maybe use 
eps_t -> v
eps_r -> w
to make this easier to parse.

Sources/SwiftFusion/Geometry/Pose2.swift
```diff
@@ -1,34 +1,223 @@
...
+    rotStorage = r
+  }
+
+  /// The derivative of `init` satisfiying convention (2) described above.
+  @derivative(of: init(_:_:))
+  @usableFromInline
+  static func vjpInit(_ t: Point2, _ r: Rot2)
+    -> (value: Pose2, pullback: (TangentVector) -> (Point2.TangentVector, Rot2.TangentVector))
+  {
+    // Explanation of this calculation:
```
dellaert, line 56: awesome

Sources/SwiftFusion/Geometry/Pose2.swift
```diff
@@ -1,34 +1,223 @@
...
+    //     = init(r^-1 * exp(eps_t), exp(eps_r))
+    //
+    // Since `eps_t`, `eps_r` are small, we can take logs of both sides to get
+    //   dinit_(t, r)(eps_t, eps_r) = (r^-1 * eps_t, eps_r)
+    //
+    // Or in matrix notation:
+    //   dinit_(t, r)(eps_t, eps_r) =  ( r^-1  0 ) ( eps_t )
+    //                                 ( 0     1 ) ( eps_r )
+    //
+    // We actually need the pullback rather than the differential, so that's the transpose:
```
dellaert, line 75: makes sense !

Sources/SwiftFusion/Geometry/Pose2.swift
```diff
@@ -1,34 +1,223 @@
...
+    //   pbinit_(t, r)(eps_t, eps_r) =  ( r 0 ) ( eps_t )
+    //                                  ( 0 1 ) ( eps_r )
+    //
+    // The pullback here implements the linear map corresponding to that matrix.
+    (Pose2(t, r), { eps in
+      (
+        Point2.TangentVector(
```
dellaert, line 82: r.rotate?

Sources/SwiftFusion/Geometry/Pose2.swift
```diff
@@ -1,34 +1,223 @@
...
+    // can get the matrix for the pullback of `t` by inverting the matrix for the pullback of
+    // `init` and then taking the relevant columns.
+    //
+    // The inverse is:
+    //   ( r^-1 0 )
+    //   ( 0    1 )
+    //
+    // The below implements the linear transformation represented by the first column of this matrix.
+    return (t, { v in
+      TangentVector(
```
dellaert, line 107: r.unrotate?

Sources/SwiftFusion/Geometry/Pose2.swift
```diff
@@ -1,34 +1,223 @@
...
+        vx: rot.c * v.x + rot.s * v.y,
+        vy: -rot.s * v.x + rot.c * v.y,
+        omega: .zero
+      )
+    })
+  }
+
+  /// The derivative of `t` satisfiying convention (2) described above.
```
dellaert, line 115: `rot`

Sources/SwiftFusion/Geometry/Pose2.swift
```diff
@@ -1,34 +1,223 @@
...
+  @derivative(of: rot)
+  @usableFromInline
+  func vjpRot() -> (value: Rot2, pullback: (Rot2.TangentVector) -> TangentVector) {
+    // This pullback is similar to the pullback in `vjpT`, but for the second column of the
+    // matrix. See the comment in `vjpT` for more information.
+    return (rot, { TangentVector(vx: 0, vy: 0, omega: $0) })
+  }
+
+  /// The Lie algebra se(2).
+  public struct TangentVector:
```
dellaert, line 125: Yeah, it's beautiful to have these, but again: Vector3 will do fine for our purposes, and gives us vector space properties for free. (and they are easy to concatenate them later into TangentVectors for product manifolds)

Sources/SwiftFusion/Geometry/Pose2.swift
```diff
@@ -1,34 +1,223 @@
...
-    rot_ = r
+  public init(_ x: Double, _ y: Double, _ theta: Double) {
+    self.init(Point2(x, y), Rot2(theta))
   }
 
-  public static func == (lhs: Pose2, rhs: Pose2) -> Bool {
-    (lhs.t_, lhs.rot_) == (rhs.t_, rhs.rot_)
+  public init(randomWithCovariance covariance: Tensor<Double>) {
+    self.init(0, 0, 0)
+    let r = matmul(cholesky(covariance), Tensor<Double>(randomNormal: [3, 1])).scalars
```
dellaert, line 182: awesome. In later PR this will probably migrate out :-)

Sources/SwiftFusion/Geometry/Pose2.swift
```diff
@@ -1,34 +1,223 @@
...
-  public var t_: Point2
-  public var rot_: Rot2
+import TensorFlow
 
+/// SE(2) Lie group of 2D Euclidean Poses.
+///
+/// We adopt the conventions in [1] and [2] for the tangent space of this manifold and for
+/// derivatives of functions to/from this manifold. In particular:
+///
+///   1. The tangent space at every point is the Lie algebra so(2).
```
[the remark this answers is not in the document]
marcrasi: Whoops, typo, I did mean se(2).

Your argument about making `TangentVector == Vector3` is convincing. I'll change it. This will require me to actually create the `Vector*` types, and I would like to do that in a separate PR so that this one doesn't get way too big and complicated, so I'll work on creating and merging that before updating the PR to use it.

(I'm particularly convinced by the thing you said during our video chat about how writing a data structure for a linearized factor graph is easy when you make all of the tangent vectors of all the different types be compatible "flat" vectors. I didn't really see the connection between this and that until I actually started playing with a "dumb" implementation of a NonlinearFactorGraph, but now I see :))

Sources/SwiftFusion/Geometry/Pose2.swift
marcrasi: ah yes, "easy to concatenate them later into TangentVectors for product manifolds" is the thing that makes it easy to linearize factor graphs

Sources/SwiftFusion/Geometry/Pose2.swift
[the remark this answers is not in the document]
dellaert: Awesome to be on the same page - it will also allow to easily port many unit tests from GTSAM. But the Pose2 exercise shows how awesome the synthesized derivatives approach is!!! I am very impressed by that.

Sources/SwiftFusion/Geometry/Pose2.swift
```diff
@@ -1,34 +1,223 @@
...
+    /// https://github.com/borglab/gtsam/blob/develop/doc/LieGroups.pdf.
+    public var vx: Double
+
+    /// The y translation component in the basis defined in section 4.1 of
+    /// https://github.com/borglab/gtsam/blob/develop/doc/LieGroups.pdf.
+    public var vy: Double
+
+    /// The rotation component in the basis defined in section 4.1 of
+    /// https://github.com/borglab/gtsam/blob/develop/doc/LieGroups.pdf.
+    public var omega: Double
```
ProfFan, line 140: I think this can be `w`?

ProfFan:
This is great! Having actual matrices as tangent vectors would unblock a lot of stuff :)

Tests/SwiftFusionTests/TestUtilities.swift
```diff
@@ -0,0 +1,23 @@
...
+import XCTest
+
+/// Asserts that `x` and `y` have the same shape and that their values have absolute difference
+/// less than `accuracy`.
+func assertEqual<T: TensorFlowFloatingPoint>(
+  _ x: Tensor<T>, _ y: Tensor<T>, accuracy: T, file: StaticString = #file, line: UInt = #line
+) {
+  guard x.shape == y.shape else {
```
ProfFan, line 9: Can this be picked out and merged first? :)

dellaert: Nothing is stopping you :-) - new PR with cherry-pick or just the copied code

marcrasi: Yes, please go ahead if you want. Or if you'd like me to do it, I could send out a PR sometime tonight. It'll probably be a few hours.

ProfFan: I think that will create a merge conflict here, but I will try :)

Sources/SwiftFusion/Geometry/Pose2.swift
```diff
@@ -1,34 +1,223 @@
...
+
+  /// Creates a `Pose2` with rotation `r` and translation `t`.
+  ///
+  /// This is the bijection SO(2) x R^2 -> SE(2), where "x" means direct product of groups. (Note:
+  /// not a group homomorphism!)
   @differentiable
-  public init(_ x: Double, _ y: Double, _ theta: Double) {
-    t_ = Point2(x, y)
-    rot_ = Rot2(theta)
+  public init(_ t: Point2, _ r: Rot2) {
```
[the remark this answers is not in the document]
marcrasi: Let's do (2). I like the idea of making the order in Pose2 and Pose3 consistent with each other. I guess this will make all our Pose2 vectors, jacobians, etc, inconsistent with the GTSAM ones, which is unfortunate.

I'll update this PR to do (2).

marcrasi:
I've updated the TangetVector to use Vector3, and I've made all the suggested changes in the names and doc comments. Please take another look when you have time :)

The adjoint is currently named `groupAdjoint` because that's what it seemed like we were leaning towards in https://github.com/borglab/SwiftFusion/issues/21 when I updated that part of the code. I'd be happy to change it to `Adjoint` if we eventually land on that. Maybe we can make that change after merging this PR, though, so that this PR isn't blocked by a long naming discussion :)

PR #19, opened 2020-04-02
add Vector1, Vector2, Vector3 types

### LINKED ISSUE #17: Vector types
It would be nice to have some `Vector2`, `Vector3`, etc types.

We can set their tangent spaces equal to themselves.

We can replace `Point2` with the `Vector2` type.

marcrasi:
* Adds Vector1, Vector2, Vector3 types using generated code.
  - There's no way to add a "generate" step to a SwiftPM build, so the generated code is checked in next to the templates and there is a script for regenerating it.
* Adds methods for converting to/from `Tensor` so that we can use `TensorFlow` linear algebra methods.
* Changes `Rot2.TangentVector` to `Vector1`.
* Removes `Point2` and uses `Vector2` everywhere instead.
* Did not change `Pose2.TangentVector` to use these yet, because I'll do that in #16 after this is merged.

Reason: We want some common types for tangent vectors of things, so that it's easy to concatenate them together and do linear algebra on them.

I implemented the vector type storage as structs of `Double`, which I think should be faster than implementing the storage as `Tensor`, because `Tensor` involves a bunch of metadata, indirection, and reference counting. I think it'll still be faster even if we occasionally convert to/from `Tensor` in order to use `TensorFlow` functionaility. We can always swap the storage implementation later once we start measuring stuff and figuring out what's actually fastest.

Resolves #17.


Sources/SwiftFusion/Core/Vector.swift
```diff
@@ -0,0 +1,285 @@
+// WARNING: This is a generated file. Do not edit it. Instead, edit the corresponding ".gyb" file.
+// See "generate.sh" in the root of this repository for instructions how to regenerate files.
+
+// ###sourceLocation(file: "/usr/local/google/home/marcrasi/git/SwiftFusion/Sources/SwiftFusion/Core/Vector.swift.gyb", line: 1)
```
ProfFan, line 4: Can this be made relative the the project root?

Sources/SwiftFusion/Core/Vector.swift.gyb
```diff
@@ -0,0 +1,84 @@
...
+}
+
+/// Normed vector space methods.
+extension Vector${dim}: VectorProtocol {
+  public typealias VectorSpaceScalar = Double
+
+  @differentiable
+  public var magnitude: Double {
+    var squaredMagnitude: Double = 0
+    % for coordinate in coordinates:
```
ProfFan, line 30: maybe `self.squared().sum()`?

Sources/SwiftFusion/Core/Vector.swift.gyb
```diff
@@ -0,0 +1,84 @@
...
+    % end
+  }
+}
+
+/// Normed vector space methods.
+extension Vector${dim}: VectorProtocol {
+  public typealias VectorSpaceScalar = Double
+
+  @differentiable
+  public var magnitude: Double {
```
ProfFan, line 28: maybe `norm`

Sources/SwiftFusion/Core/Vector.swift.gyb
```diff
@@ -0,0 +1,84 @@
+import TensorFlow
+
+% dims = [1, 2, 3]
+% for dim in dims:
+% coordinates = ['x', 'y', 'z'][0:dim]
+
+/// An element of R^${dim}.
+public struct Vector${dim}:
+  AdditiveArithmetic, Differentiable, ElementaryFunctions, KeyPathIterable, TangentStandardBasis
```
ProfFan, line 9: I see what you are trying to achieve here, but can we have something that is directly compatible with a TF Tensor?

Sources/SwiftFusion/Geometry/Pose2.swift
```diff
@@ -28,7 +28,7 @@ public struct Pose2: Equatable, Differentiable, KeyPathIterable, TangentStandard
 
 @differentiable
 public func inverse(_ p: Pose2) -> Pose2 {
-  Pose2(inverse(p.rot_), p.rot_.unrotate(-p.t_))
+  Pose2(inverse(p.rot_), p.rot_.unrotated(-p.t_))
```
ProfFan, line 31: I think this should be `unrotate` because `p.rot_` is not modified.

generate.sh
```diff
@@ -0,0 +1,33 @@
...
+#   3. Run `./generate $SWIFT_BASE`.
+
+SWIFT_BASE="$1"
+if [ -z "$SWIFT_BASE" ]
+then
+  echo "SWIFT_BASE not specified. See instructions at the top of generate.sh."
+  exit 1
+fi
+
+GYB="$SWIFT_BASE"/utils/gyb
```
ProfFan, line 17: This is my first time seeing a GYB haha :)

Sources/SwiftFusion/Core/Vector.swift.gyb
dellaert: SquaredNorm

Sources/SwiftFusion/Geometry/Pose2.swift
dellaert: Agreed, plus matches GTSAM

Sources/SwiftFusion/Core/Vector.swift.gyb
marcrasi: This is as close as we can get with a wrapper type -- you can call `.tensor` to get the TF Tensor and you can do `VectorN(t)` to convert a TF Tensor to one of these.

We could use a simple unwrapped TF Tensor. I like the static type information about the dimension that we get from the wrapper, but I wouldn't object to just using TF Tensor if you think that would be better.

dellaert: I like the small vector types - we also have this in GTSAM with static Eigen vectors. I think it is key to performance, and the static type checking is great. It does add extra code and complexity...

marcrasi: I had a somewhat related thought when thinking about this.

We could allow a "batch dimension" in these vectors so that for example a `Vector2` is actually a `2 x batchSize`. Then, if we write all operations to distribute over the batch dimension correctly, we can pack all the vectors that we're operating on together and do the operations on all the vectors at the same time. This may speed things up a lot.

What do you think? Does GTSAM do something similar anywhere?

ProfFan: Yea I agree on that :)

dellaert: KISS :-) feels like premature optimization at this point?

dellaert: I’m totally open to these type of things but I’d focus on the mission now. Which is to get Gauss Newton and CGLS going on PoseSLAM :-) we can then measure...

marcrasi:
Thanks for the comments! All addressed now.

PR #20, opened 2020-04-03
Remove SwiftRT

ProfFan:
Let's not over-engineer and stick with TensorFlow for now :)

~~Mergeable after the test utility function `assertEqual` in #16 is merged.~~
I copied that piece of code from Marc's PR, now mergeable.

This should also reduce compile time from 1 minute to about 10 seconds.


dellaert:
I don’t even know what Swift RT is, so I’ll leave you two to sort this out :-)

PR #24, opened 2020-04-03
draft simple inflexible factor graph api

marcrasi:
I've been playing with a factor graph API. Here's a super unpolished, untested, inflexible API.

Having something concrete like this could help us discuss concretely how the API should be done.

### DIFF
--- a/Sources/SwiftFusion/Nonlinear/NonlinearFactorGraph.swift
+++ b/Sources/SwiftFusion/Nonlinear/NonlinearFactorGraph.swift
@@ -1,0 +1,114 @@
+import TensorFlow
+
+public struct NonlinearFactorGraph {
+  var factor1s: [Factor1] = []
+  var factor2s: [Factor2] = []
+
+  var factorCount: Int { factor1s.count + factor2s.count }
+
+  public init() {}
+
+  public mutating func add(
+    _ value1Index: Int, _ error: @escaping @differentiable (Pose2) -> Pose2, _ noise: NoiseModel
+  ) {
+    factor1s.append(Factor1(value1Index: value1Index, error: error, noise: noise))
+  }
+
+  public mutating func add(
+    _ value1Index: Int, _ value2Index: Int,
+    _ error: @escaping @differentiable (Pose2, Pose2) -> Pose2,
+    _ noise: NoiseModel
+  ) {
+    factor2s.append(Factor2(
+      value1Index: value1Index, value2Index: value2Index, error: error, noise: noise))
+  }
+
+  public func loss(at point: [Pose2]) -> Double {
+    factor1s.map { $0.noise.loss($0.error(point[$0.value1Index])) }.reduce(0, +) +
+      factor2s.map { $0.noise.loss($0.error(point[$0.value1Index], point[$0.value2Index])) }.reduce(0, +)
+  }
+
+  /// Returns an nx1 matrix (a vector) whose entries are the factors' errors at `point`.
+  public func errors(at point: [Pose2]) -> Tensor<Double> {
+    var vector = Tensor<Double>(zeros: [3 * factorCount, 1])
+    var currentFactorIndex: Int = 0
+    for factor in factor1s {
+      let error = factor.error(point[factor.value1Index])
+      vector[3 * currentFactorIndex, 0] = Tensor(error.rot.theta)
+      vector[3 * currentFactorIndex + 1, 0] = Tensor(error.t.x)
+      vector[3 * currentFactorIndex + 2, 0] = Tensor(error.t.y)
+      currentFactorIndex += 1
+    }
+    for factor in factor2s {
+      let error = factor.error(point[factor.value1Index], point[factor.value2Index])
+      vector[3 * currentFactorIndex, 0] = Tensor(error.rot.theta)
+      vector[3 * currentFactorIndex + 1, 0] = Tensor(error.t.x)
+      vector[3 * currentFactorIndex + 2, 0] = Tensor(error.t.y)
+      currentFactorIndex += 1
+    }
+    return vector
+  }
+
+  /// Returns a matrix representing a linear approximation of changes in factor errors as a
+  /// function of changes in `point`.
+  ///
+  /// Note: Currently returns a dense representation of the matrix, even though the matrix values
+  /// are actually very sparse.
+  public func linearize(at point: [Pose2]) -> Tensor<Double> {
+    var matrix = Tensor<Double>(zeros: [3 * factorCount, 3 * point.count])
+    var currentFactorIndex: Int = 0
+    for factor in factor1s {
+      let pb = pullback(at: point[factor.value1Index], in: factor.error)
+      for (basisVectorIndex, basisVector) in Pose2.tangentStandardBasis.enumerated() {
+        let rowValue = pb(basisVector)
+        let rowIndex = 3 * currentFactorIndex + basisVectorIndex
+        matrix[rowIndex, 3 * factor.value1Index] = Tensor(rowValue.x)
+        matrix[rowIndex, 3 * factor.value1Index + 1] = Tensor(rowValue.y)
+        matrix[rowIndex, 3 * factor.value1Index + 2] = Tensor(rowValue.z)
+      }
+      currentFactorIndex += 1
+    }
+    for factor in factor2s {
+      let pb = pullback(at: point[factor.value1Index], point[factor.value2Index], in: factor.error)
+      for (basisVectorIndex, basisVector) in Pose2.tangentStandardBasis.enumerated() {
+        let (rowValue1, rowValue2) = pb(basisVector)
+        let rowIndex = 3 * currentFactorIndex + basisVectorIndex
+        matrix[rowIndex, 3 * factor.value1Index] = Tensor(rowValue1.x)
+        matrix[rowIndex, 3 * factor.value1Index + 1] = Tensor(rowValue1.y)
+        matrix[rowIndex, 3 * factor.value1Index + 2] = Tensor(rowValue1.z)
+        matrix[rowIndex, 3 * factor.value2Index] = Tensor(rowValue2.x)
+        matrix[rowIndex, 3 * factor.value2Index + 1] = Tensor(rowValue2.y)
+        matrix[rowIndex, 3 * factor.value2Index + 2] = Tensor(rowValue2.z)
+      }
+      currentFactorIndex += 1
+    }
+    return matrix
+  }
+}
+
+public struct NoiseModel {
+  public let sigmaX, sigmaY, sigmaRot: Double
+  public func loss(_ error: Pose2) -> Double {
+    sigmaX * error.t.x * error.t.x +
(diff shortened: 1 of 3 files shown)


Sources/SwiftFusion/Nonlinear/NonlinearFactorGraph.swift
```diff
@@ -0,0 +1,114 @@
+import TensorFlow
+
+public struct NonlinearFactorGraph {
+  var factor1s: [Factor1] = []
+  var factor2s: [Factor2] = []
```
marcrasi, line 5: Not polymorphic over the variable types yet. Not even polymorphic over the arities of the factors.

Sources/SwiftFusion/Nonlinear/NonlinearFactorGraph.swift
```diff
@@ -0,0 +1,114 @@
...
+      factor2s.map { $0.noise.loss($0.error(point[$0.value1Index], point[$0.value2Index])) }.reduce(0, +)
+  }
+
+  /// Returns an nx1 matrix (a vector) whose entries are the factors' errors at `point`.
+  public func errors(at point: [Pose2]) -> Tensor<Double> {
+    var vector = Tensor<Double>(zeros: [3 * factorCount, 1])
+    var currentFactorIndex: Int = 0
+    for factor in factor1s {
+      let error = factor.error(point[factor.value1Index])
+      vector[3 * currentFactorIndex, 0] = Tensor(error.rot.theta)
```
marcrasi, line 37: Not using any of the Vector -> Tensor conversion functions yet.

Sources/SwiftFusion/Nonlinear/NonlinearFactorGraph.swift
```diff
@@ -0,0 +1,114 @@
+import TensorFlow
```
dellaert, line 1: Comment

Sources/SwiftFusion/Nonlinear/NonlinearFactorGraph.swift
```diff
@@ -0,0 +1,114 @@
...
+
+public struct NonlinearFactorGraph {
```
dellaert, line 3: comment

Sources/SwiftFusion/Nonlinear/NonlinearFactorGraph.swift
```diff
@@ -0,0 +1,114 @@
...
+  var factor1s: [Factor1] = []
```
dellaert, line 4: I think all we need is be polymorphic about factor type. But maybe this "old thinking" to have a abstract base class? What the new way? Protocol?

Sources/SwiftFusion/Nonlinear/NonlinearFactorGraph.swift
dellaert: I think the variable types are only defined/needed in the factors - not here.

Sources/SwiftFusion/Nonlinear/NonlinearFactorGraph.swift
```diff
@@ -0,0 +1,114 @@
...
+  var factor2s: [Factor2] = []
+
+  var factorCount: Int { factor1s.count + factor2s.count }
+
+  public init() {}
+
+  public mutating func add(
```
dellaert, line 11: Here I think we just need `add(factor)`

Sources/SwiftFusion/Nonlinear/NonlinearFactorGraph.swift
```diff
@@ -0,0 +1,114 @@
...
+    _ value1Index: Int, _ value2Index: Int,
+    _ error: @escaping @differentiable (Pose2, Pose2) -> Pose2,
+    _ noise: NoiseModel
+  ) {
+    factor2s.append(Factor2(
+      value1Index: value1Index, value2Index: value2Index, error: error, noise: noise))
+  }
+
+  public func loss(at point: [Pose2]) -> Double {
+    factor1s.map { $0.noise.loss($0.error(point[$0.value1Index])) }.reduce(0, +) +
```
dellaert, line 27: just one set of factors, but this looks good, if a bit concise :-) I think it might look better with explicit arguments in the lambda. "Typing is not the bottleneck"

Sources/SwiftFusion/Nonlinear/NonlinearFactorGraph.swift
```diff
@@ -0,0 +1,114 @@
...
+      factor2s.map { $0.noise.loss($0.error(point[$0.value1Index], point[$0.value2Index])) }.reduce(0, +)
+  }
+
+  /// Returns an nx1 matrix (a vector) whose entries are the factors' errors at `point`.
+  public func errors(at point: [Pose2]) -> Tensor<Double> {
```
dellaert, line 32: should this function be called above? And, rather than one big tensor, it's just a collection of multivariate errors, one per factor

Sources/SwiftFusion/Nonlinear/NonlinearFactorGraph.swift
dellaert: I think we might never go there: just insist a factor returns a tensor?

Sources/SwiftFusion/Nonlinear/NonlinearFactorGraph.swift
```diff
@@ -0,0 +1,114 @@
...
+  public mutating func add(
+    _ value1Index: Int, _ value2Index: Int,
+    _ error: @escaping @differentiable (Pose2, Pose2) -> Pose2,
+    _ noise: NoiseModel
+  ) {
+    factor2s.append(Factor2(
+      value1Index: value1Index, value2Index: value2Index, error: error, noise: noise))
+  }
+
+  public func loss(at point: [Pose2]) -> Double {
```
dellaert, line 26: needs to take a polymorphic `Values` ?

Sources/SwiftFusion/Nonlinear/NonlinearFactorGraph.swift
```diff
@@ -0,0 +1,114 @@
...
+    }
+    return vector
+  }
+
+  /// Returns a matrix representing a linear approximation of changes in factor errors as a
+  /// function of changes in `point`.
+  ///
+  /// Note: Currently returns a dense representation of the matrix, even though the matrix values
+  /// are actually very sparse.
+  public func linearize(at point: [Pose2]) -> Tensor<Double> {
```
dellaert, line 57: Yeah, dense will never be needed :-) So, I think it is as with above: just call linearize on each factor to return a JacobianFactor, which would be the GTSAM way. *But* with Fan we are debating whether we even need that: we just need vjp and jvp defined for each factor to implement CG. I think we need to be TDD here :-)

ProfFan:
@marcrasi Can we have a meeting over hangouts at around 5:45 PM EDT?

Sources/SwiftFusion/Nonlinear/NonlinearFactorGraph.swift
ProfFan: I think we can have a `Factor` protocol that defines a loss function, and a `DifferentiableFactor` with both loss and  derivative.

Sources/SwiftFusion/Nonlinear/NonlinearFactorGraph.swift
marcrasi: Oh yes, that's a good idea.

Sources/SwiftFusion/Nonlinear/NonlinearFactorGraph.swift
marcrasi: Ideally we would not use classes because they do not interact with AD very well:
 * the AD implementation for classes isn't very robust or fleshed out yet, and
 * even if the implementation was robust, AD is probably always going to behave in weird unexpected ways in the presence of classes because classes have [reference semantics](https://developer.apple.com/swift/blog/?id=10) and it's not obvious what it means to differentiate functions involving references to things.

Also, structs are just generally more encouraged than classes in Swift, so it's good to use them when possible.

We should indeed be able to do this with some kind of `Factor` protocol and structs that conform to that protocol. (Why a separate `DifferentiableFactor` protocol? Can we just make `Factor` have all the necessary stuff for differentiation?) Once we have a `Factor` protocol, we can store an array `[Factor]` in the factor graph. (The type `Factor` is called an "existential" and it can store values of any type that conforms to `Factor`).

There is one difficulty with this approach. The obvious `Factor` protocol would be something like this:
```
protocol Factor {
  associatedtype InputType: Differentiable

  @differentiable
  func error(_ input: InputType) -> Tensor<Double>
}
```

Unfortunately, Swift does not support existentials for protocols with `associatedtype`s, so we can't store an array of those in the factor graph.

We can work around this by defining something like:

```
protocol Factor {
  func error(anyInput: Any) -> Tensor<Double>
  func errorPullback(anyInput: Any, _ v: Tensor<Double>) -> Tensor<Double>
}
```

This has no associated types, so we can use its existential.

Then we can define helper protocols so that users defining factors do not need to do all the casting themselves:

```
protocol TypedFactor1 {
  associatedtype InputType: Differentiable
  @differentiable func error(_ input: InputType) -> Tensor<Float>
}

extension TypedFactor1 {
  func error(anyInput: Any) -> Tensor<Double> {
    let input = anyInput as! InputType
    return error(input)
  }

  func errorPullback(anyInput: Any) -> (Tensor<Double>) -> Tensor<Double> {
    let input = anyInput as! InputType
    return pullback(at: input, in: { error($0) })
  }
}

struct PriorFactor: TypedFactor1, Factor {
  let prior: Pose2

  @differentiable
  func error(_ input: Pose2) -> Tensor<Float> {
    between(input, prior).tensor
  }
}
```

I think this is probably the part of the API that I can be most helpful on, so I'm going to play with this and see if I can get a working prototype to show you before we have a call.

marcrasi:
Sorry that I didn't notice your question about a hangouts meeting until pretty late! We should definitely have a hangouts some time. 11am EDT (9am PDT) Monday would work for me. I sent a google calendar invite to @dellaert for that time (I don't know if you have an email that works with google calendar @ProfFan, let me know if you do, and I can send the invite to you too). Of course feel free to move the event around if a different time would be better for you.

Before we talk, I'm going to try to prototype a `Factor` protocol like I discussed in my inline comment.

marcrasi:
Oops, I meant "noon EDT (9am PDT)".

Sources/SwiftFusion/Nonlinear/NonlinearFactorGraph.swift
dellaert: Yes, that sounds great!
Note there is another, more ad-hoc way in GTSAM to create factors which is "AddExpressionFactor", where you give an expression (created in a compile time DSL), a measurement, and a noise model. The key feature there is to have "variable expressions" that refer to the Values through a key - very much  like tensorflow "Variables"  in fact. Obvious question: can we hijack the machinery in swift for tensorflow almost directly to support that style?

PR #26, opened 2020-04-06
Remove redundant code without test coverage

ProfFan:
Remove redundant code without test coverage


ProfFan:
This `Between` is what I added in the experimenting phase and is never covered in the test suite.

PR #29, opened 2020-04-10
make Dictionary Differentiable

marcrasi:
As discussed in https://github.com/borglab/SwiftFusion/pull/23/files/eea06d0b6c3fca7c6bba2ebe6588097f0096dc36#r405894624, it would be nice for `Dictionary` to be `Differentiable`.

I will work on getting this included in the standard library so that it does not need to exist in this repo, but it will take weeks-months for that to make it into a release toolchain, so in the meantime you probably want it in this repo so that you can use it sooner.

This PR only defines a derivative for the "subscript" operation. We'll need to define derivatives for any other primitive dictionary operations that you need. But it looks like "subscript" is all you need in the short term.

### DIFF
--- a/Sources/SwiftFusion/Core/Dictionary+Differentiable.swift
+++ b/Sources/SwiftFusion/Core/Dictionary+Differentiable.swift
@@ -1,0 +1,52 @@
+/// This file makes `Dictionary` differentiable.
+///
+/// Note: This will eventually be moved into the Swift standard library. Once it is in the
+/// standard library, we can delete it from this repository.
+
+/// Implements the `Differentiable` requirements.
+extension Dictionary: Differentiable where Value: Differentiable {
+  public typealias TangentVector = Dictionary<Key, Value.TangentVector>
+  public mutating func move(along direction: TangentVector) {
+    for (componentKey, componentDirection) in direction {
+      func fatalMissingComponent() -> Value {
+        fatalError("missing component \(componentKey) in moved Dictionary")
+      }
+      self[componentKey, default: fatalMissingComponent()].move(along: componentDirection)
+    }
+  }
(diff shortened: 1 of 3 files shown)


ProfFan:
Wow! This is fast!

Sources/SwiftFusion/Core/Dictionary+Differentiable.swift
```diff
@@ -0,0 +1,52 @@
...
+///
+/// Note: Once the differentiable `Dictionary` is moved into the standard library, the standard
+/// `Dictionary` methods will be differentiable and you won't have to use special differentiable
+/// methods to manipulate `Dictionary`.
+extension Dictionary where Value: Differentiable {
+  /// Returns the value with `key`.
+  ///
+  /// Precondition: `self` contains an entry with key `key`.
+  @differentiable
+  public func differentiableSubscript(_ key: Key) -> Value {
```
ProfFan, line 41: So the user need to call this instead of `dict[ key ]` (the normal subscript), right?

dellaert:
indeed :-) Reading through this now...

Sources/SwiftFusion/Core/Dictionary+Differentiable.swift
marcrasi: Yes. The compiler should give nice errors pointing at any uses of `dict[ key ]` in differentiable functions, so it should be pretty easy to fix whenever the user forgets.

Eventually we'll be able to support derivatives of `dict[ key ]`, but there are a few things blocking them from working right now.

marcrasi:
> Will this be KeyPathIterable to Double?

Yes, though there might be some problems with nondeterministic ordering of the key paths depending on what you need them for. It shouldn't be too hard to make the ordering deterministic (e.g. by a given key order) if you do need that.

PR #33, opened 2020-04-15
import AnyDifferentiable

### LINKED ISSUE #30: Type-erased wrapper for differentiable `Value` (a.k.a. `AnyDifferentiable`)
I think @marcrasi has mentioned this in an issue, and I encountered this problem in prototyping the nonlinear factor graph API as well.

marcrasi:
Adds an `AnyDifferentiable` type that allows us to take derivatives with respect to type-erased differentiable types.

Includes a test that demonstrates how to take the derivative with respect to an array that contains `Pose2` and `Vector2` elements.

NOTE: This will be available in S4TF v0.9, so we should delete this as soon as we switch to that version.

Resolves #30


ProfFan:
Thanks :)

PR #34, opened 2020-04-22
Add Nonlinear FG interface

ProfFan:
This adds a nonlinear factor graph interface for SwiftFusion.

Currently only implemented BetweenFactor for Pose2.

TODO:
- [ ] `AnyComposable` for generalizing `BetweenFactor` to `Rot2`, etc.
- [x] Test linear CG on linearize FG


ProfFan:
@marcrasi I blatantly copied the `AnyDifferentiable` code to make `AnyNonlinearFactor`... I think some of the code is redundant but I am not sure. Also, do you have any suggestions for the API?

@dellaert Since the `CGLS` optimizer is working on the linearized factor graph, I think this PR is ready for an initial review :)

ProfFan:
BTW, for Pose2SLAM demo the solution converges applying CGLS two times, which is amazing!

dellaert:
Very cool. Have a full day of meetings but hope to get to this before our 1:1.  Amazing that the code works but convergence is not amazing if you start close to the optimum :-)

Sources/SwiftFusion/Inference/BetweenFactor.swift
```diff
@@ -0,0 +1,89 @@
...
+  }
+  
+  @differentiable(wrt: values)
+  public func errorVector(_ values: Values) -> Vector3 {
+    let error = between(
+      between(values[keys[1]].baseAs(Pose2.self), values[keys[0]].baseAs(Pose2.self)),
+      difference
+    )
+    
+    return Vector3(error.rot.theta, error.t.x, error.t.y)
```
marcrasi, line 77: This should actually return something using local coordinates (`between(values[key1].baseAs(Pose2.self), values[key2].baseAs(Pose2.self).localCoordinates(around: difference)`), right?

If so, you could add a TODO and a github issue for this, because this seems to work for now.

Sources/SwiftFusion/Inference/Factor.swift
```diff
@@ -30,11 +30,33 @@ public protocol Factor {
...
+/// A `NonlinearFactor` corresponds to the `NonlinearFactor` in GTSAM.
+///
+/// Input is a dictionary of `Key` to `Value` pairs, and the output is the scalar
+/// error value
+///
+/// Interpretation
+/// ================
+/// `Input`: the input values as key-value pairs
+///
+public protocol NonlinearFactor: Differentiable & Factor {
```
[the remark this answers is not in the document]
ProfFan: Yeah I agree on the first one, but we will still need the type-erased wrapper if we have associated types in NonlinearFactor, right?

@dellaert What's your stance on this (pt. 2)?

Sources/SwiftFusion/Inference/Factor.swift
```diff
@@ -15,7 +15,7 @@ import TensorFlow
 
 /// The most general factor protocol.
 public protocol Factor {
-  var keys: Array<Int> { get set }
+  var keys: Array<Int> { get }
```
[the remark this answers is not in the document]
ProfFan: I think we still have this in some of the algorithms for iterating through the keys to lookup the Values, but that may change if we have a better encapsulated API

Sources/SwiftFusion/Inference/BetweenFactor.swift
```diff
@@ -0,0 +1,89 @@
...
+  }
+  
+  public func linearize(_ values: Values) -> JacobianFactor {
```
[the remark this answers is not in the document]
ProfFan: Nice!

Sources/SwiftFusion/Inference/BetweenFactor.swift
```diff
@@ -0,0 +1,89 @@
...
+/// Input is a dictionary of `Key` to `Value` pairs, and the output is the scalar
+/// error value
+///
+/// Interpretation
+/// ================
+/// `Input`: the input values as key-value pairs
+///
+public struct BetweenFactor: NonlinearFactor {
+  @noDerivative
+  public var keys: Array<Int> = []
```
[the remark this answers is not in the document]
ProfFan: Yeah this is me simply being lazy :)

Sources/SwiftFusion/Inference/BetweenFactor.swift
ProfFan: Correct, this is on my TODO :)

Will change shortly...

Sources/SwiftFusion/Inference/Factor.swift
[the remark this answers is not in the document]
marcrasi: Yes, we will need the type-erased wrapper if there are associated types in NonlinearFactor. But currently there aren't any associated types (other than the `TangentVector` that comes from `Differentiable`), and I can't think of any associated types that you will need in the future.

Tests/SwiftFusionTests/Inference/NonlinearFactorGraphTests.swift
```diff
@@ -0,0 +1,99 @@
...
+      
+      var dx = VectorValues()
+      
+      for i in 0..<5 {
+        dx.insert(i, Tensor<Double>(shape: [3, 1], scalars: [0, 0, 0]))
+      }
+      
+      optimizer.optimize(gfg: gfg, initial: &dx)
+      
+      for i in 0..<5 {
```
marcrasi, line 70: I see two possible improvements to the `Values` API.

1. You could make a subscript that does casting, which would allow you to mutate variables without storing them in an intermediate `var`. Something like `val[i, as: Pose2.self].move(along: ...)`.
2. `Values.TangentVector` could be `VectorValues`, which would allow you to replace this whole loop with a single line `val.move(along: dx)`.

Let me know if you want me to help write code for either of these.

dellaert:
Will review now...

Sources/SwiftFusion/Inference/AnyNonlinearFactor.swift
```diff
@@ -0,0 +1,95 @@
...
+
+  /// The underlying base value, type-erased to `Any`.
+  var _typeErasedBase: Any { get }
+
+  /// Returns the underlying value unboxed to the given type, if possible.
+  func _unboxed<U: NonlinearFactor>(to type: U.Type) -> U?
+  
+  var _keys: Array<Int> { get }
+  
+  @differentiable(wrt: values)
```
dellaert, line 16: document all - expectations

Sources/SwiftFusion/Inference/AnyNonlinearFactor.swift
```diff
@@ -0,0 +1,95 @@
...
+  func _error(_ values: Values) -> ScalarType {
+    _base.error(values)
+  }
+  
+  func _linearize(_ values: Values) -> JacobianFactor {
+    _base.linearize(values)
+  }
+}
+
+public struct AnyNonlinearFactor: NonlinearFactor {
```
dellaert, line 56: document class

Sources/SwiftFusion/Inference/AnyNonlinearFactor.swift
```diff
@@ -0,0 +1,95 @@
...
+    return _box._typeErasedBase
+  }
+
+  /// Creates a type-erased derivative from the given derivative.
+  public init<T: NonlinearFactor>(_ base: T) {
+    self._box = _ConcreteNonlinearFactorBox<T>(base)
+  }
+}
+
+extension AnyNonlinearFactor {
```
dellaert, line 74: why are some things done as extensions and others not?

Sources/SwiftFusion/Inference/BetweenFactor.swift
```diff
@@ -0,0 +1,89 @@
...
+//  static func _vjpDot<T: Differentiable & KeyPathIterable>(_ a: T, _ b: T) -> (
+//    value: Double,
+//    pullback: (Double) -> (T.TangentVector, T.TangentVector)
+//  ) {
+//    return (value: dot(a, b), pullback: { v in
+//      ((at.scaled(by: v), bt.scaled(by: v)))
+//    })
+//  }
+  
+  /// Returns the `error` of the factor.
```
dellaert, line 59: In GTSAM this is decided by the noise model - we should do the same here I think to allow for robust error models

Sources/SwiftFusion/Inference/Factor.swift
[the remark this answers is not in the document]
dellaert: We will need it for direct solvers, to decide on the ordering.

Tests/SwiftFusionTests/Inference/NonlinearFactorGraphTests.swift
```diff
@@ -0,0 +1,99 @@
...
+    
+    let expected = Tensor<Double>(shape:[3, 1], scalars: [.pi, 0.0, 0.0])
+    
+    print("gfg = \(gfg)")
+    print("error = \(gfg.residual(vv).norm)")
+    assertEqual((gfg.residual(vv))[0], expected, accuracy: 1e-6)
+  }
+  
+  /// test CGLS iterative solver
+  func testCGLSPose2SLAM() {
```
dellaert, line 33: nice!

Tests/SwiftFusionTests/Inference/NonlinearFactorGraphTests.swift
```diff
@@ -0,0 +1,99 @@
...
+
+    let map = [p1T0, p2T0, p3T0, p4T0, p5T0]
+
+    var fg = NonlinearFactorGraph()
+    
+    fg += AnyNonlinearFactor(BetweenFactor(1, 0, Pose2(2.0, 0.0, .pi / 2)))
+    fg += AnyNonlinearFactor(BetweenFactor(2, 1, Pose2(2.0, 0.0, .pi / 2)))
+    fg += AnyNonlinearFactor(BetweenFactor(3, 2, Pose2(2.0, 0.0, .pi / 2)))
+    fg += AnyNonlinearFactor(BetweenFactor(4, 3, Pose2(2.0, 0.0, .pi / 2)))
+    fg += AnyNonlinearFactor(PriorFactor(0, Pose2(0.0, 0.0, 0.0)))
```
dellaert, line 49: I'd love to be able to just say `fg += BetweenFactor(...)`

Tests/SwiftFusionTests/Inference/NonlinearFactorGraphTests.swift
dellaert: agree on both!

Tests/SwiftFusionTests/Inference/NonlinearFactorGraphTests.swift
```diff
@@ -0,0 +1,99 @@
...
+      optimizer.optimize(gfg: gfg, initial: &dx)
+      
+      for i in 0..<5 {
+        var p = val[i].baseAs(Pose2.self)
+        p.move(along: Vector3(dx[i].reshaped(toShape: [3])))
+        val[i] = AnyDifferentiable(p)
+      }
+    }
+    
+    let dumpjson = { (p: Pose2) -> String in
```
dellaert, line 77: why are we printing stuff in a test?

Sources/SwiftFusion/Inference/NonlinearFactorGraph.swift
```diff
@@ -0,0 +1,49 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+import TensorFlow
+
+/// A factor graph for nonlinear problems
+/// TODO(fan): Add noise model
+public struct NonlinearFactorGraph: FactorGraph {
+  public typealias KeysType = Array<Int>
+  
+  public typealias FactorsType = Array<AnyNonlinearFactor>
```
marcrasi, line 21: You said during our Friday meeting that you still haven't been able to eliminate `AnyNonlinearFactor` and replace this with `Array<NonlinearFactor>`.

I figured out why. The `associatedtype FactorsType : Collection where FactorsType.Element: Factor` in `protocol FactorGraph` doesn't work when you use `Array<NonlinearFactor>`, and the reason it doesn't work is that existentials can't conform to protocols. (This is a limitation of the Swift type system that probably won't go away any time soon.)

The easiest way to solve this would be to delete the `protocol FactorGraph`. Nothing actually relies on abstracting over different `FactorGraph` implementations yet, so this doesn't break anything.

I think this is worth it because deleting `AnyNonlinearFactor` makes everything a lot simpler and nicer.

It's also probably possible to redesign the `FactorGraph` protocol so that it doesn't have a constraint that breaks `Array<NonlinearFactor>`. I'd suggest writing the `FactorGraph` protocol later when we have a use case for that, because the use case will make it clearer what `protocol FactorGraph` needs to have.

Sources/SwiftFusion/Inference/AnyNonlinearFactor.swift
```diff
@@ -0,0 +1,95 @@
+//===----------------------------------------------------------------------===//
+// `AnyNonlinearFactor`
+//===----------------------------------------------------------------------===//
+
+internal protocol _AnyNonlinearFactorBox {
```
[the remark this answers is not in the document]
marcrasi: We should be able to delete this whole file if everyone agrees with my comment about deleting `AnyNonelinearFactor`.

Tests/SwiftFusionTests/Inference/NonlinearFactorGraphTests.swift
marcrasi: Deleting `AnyNonlinearFactor` would make it like that!

Sources/SwiftFusion/Inference/NonlinearFactorGraph.swift
ProfFan: @marcrasi Oh that's a great find! I'll remove the protocol for now :)

Thanks for the explanation!

marcrasi:
> AnyComposable for generalizing BetweenFactor to Rot2, etc.

I think you should be able to generalize BetweenFactor without introducing a new `Any*` type erasing type.

All you need is a protocol with the methods that `BetweenFactor` uses, and then you can use that to define a fully generic `BetweenFactor`. The protocol is likely something like `LieGroup`. Pseudocode:

```swift
protocol LieGroup {
  static func * (Self, Self) -> Self
  func inverse() -> Self
  func localCoordinates() -> LocalCoordinates
}

struct BetweenFactor<T: LieGroup> {
  var difference: T
  func error(_ value: Value) -> T.LocalCoordinates {
    let v1 = value.baseAs(T.self)
    let v2 = value.baseAs(T.self)
    return difference.localCoordinates(v2.inverse() * v1)
  }
}
```

ProfFan:
@marcrasi That means the "Manifold" type in #37 right? Maybe let's first concentrate on merging that one and then open another PR for the generic BetweenFactor, etc.

CC @dellaert

Sources/SwiftFusion/Inference/BetweenFactor.swift
ProfFan: I am thinking about adding the NoiseModel later after we got polymorphic BetweenFactor working...

Tests/SwiftFusionTests/Inference/NonlinearFactorGraphTests.swift
ProfFan: Done :)

Tests/SwiftFusionTests/Inference/NonlinearFactorGraphTests.swift
ProfFan: I think (2) will become possible after #37 is merged? Then I think we can first merge this PR and then #37, and another PR for this one.

marcrasi: I think there might be some additional work required to enable (2) after #37 is merged -- it might not be totally easy to use `Manifold` with `Values` because `Values` contains some type erased stuff and type erased stuff sometimes needs some extra work to make it play nicely with everything else.

I'll investigate how to get (2) working.

Your suggestion of fixing this in a separate PR so that we don't block this PR sounds good to me!

marcrasi:
> @marcrasi That means the "Manifold" type in #37 right?

`BetweenFactor` needs `*` and `inverse()` operations that don't exist on `Manifold`. But yes, the `localCoordinates` operation is the same as the `Manifold` local coordinates operation from #37. I think it would make sense to make a `LieGroup` protocol that refines `Manifold`. e.g.

```
protocol LieGroup: Manifold {
  static func * (Self, Self) -> Self
  func inverse() -> Self
}
```

> Maybe let's first concentrate on merging that one and then open another PR for the generic BetweenFactor, etc.

Sounds good to me!

Sources/SwiftFusion/Inference/BetweenFactor.swift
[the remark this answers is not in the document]
marcrasi: Done in https://github.com/borglab/SwiftFusion/pull/49!

Tests/SwiftFusionTests/Inference/NonlinearFactorGraphTests.swift
marcrasi: Done in https://github.com/borglab/SwiftFusion/pull/49!

ProfFan:
Closed as we have #44

PR #37, opened 2020-04-29
Recipe for defining custom manifolds

marcrasi:
This PR adds a recipe in `doc/DifferentiableManifoldRecipe.md` that describes how to define custom manifolds. To follow the recipe, all you need to do is define a global coordinate system, define a local coordinate type, and define `exp` and `log` maps that convert between them. Everything else is automatically derived.

I hope this will make it easier to define `Rot3` and `Pose3`.

This PR also:
* Adds supporting code in `Manifold.swift`.
* Rewrites `Rot2` using the recipe, to demonstrate how it works.


Sources/SwiftFusion/Geometry/Rot2.swift
```diff
@@ -1,121 +1,60 @@
...
+  public var coordinateStorage: Rot2Coordinate
+  public init(coordinateStorage: Rot2Coordinate) { self.coordinateStorage = coordinateStorage }
 
-  @usableFromInline
-  @derivative(of: c)
-  func _vjpCos() -> (value: Double, pullback: (Double) -> TangentVector) {
-    (c_, { v in Vector1(-v * self.s_) })
+  public mutating func move(along direction: Coordinate.LocalCoordinate) {
+    coordinateStorage = coordinateStorage.global(direction)
   }
```
dan-zheng, line 14: Could this definition of `move(along:)` be made a default implementation for `protocol Manifold`? It seems to use only protocol requirements.

doc/DifferentiableManifoldRecipe.md
```diff
@@ -0,0 +1,128 @@
...
+### Step 4. Create the manifold type.
+
+Create a `struct` with a single stored property `var coordinateStorage` of
+your coordinate type. Add an initializer `init(coordinateStorage:)` that
+sets the stored coordinate. Add this method to the struct:
+
+```swift
+public mutating func move(along direction: Coordinate.LocalCoordinate) {
+  coordinateStorage = coordinateStorage.global(direction)
+}
```
dan-zheng, line 83: Would this definition of `move(along:)` make sense as a default implementation on `protocol Manifold`? It's defined only in terms of `Manifold` protocol directions.

ProfFan:
This looks amazing! I will give this a shot and see if it works for Pose2, and then a review.

doc/DifferentiableManifoldRecipe.md
marcrasi: Yes in theory, but that triggered a compiler bug. I'll minimize and file an issue soon.

ProfFan:
@marcrasi I tried it on Pose2 and it works great! There is a small error though, the global update should be on the right, so it's `self * Rot2Coordinate(local.x)` not the other way around.

BTW, can you share your repo so I can push to it with the Pose2 impl?

marcrasi:
Ok, I think I gave you access to my repo!

doc/DifferentiableManifoldRecipe.md
```diff
@@ -0,0 +1,128 @@
...
+
+The coordinate `struct` you chose in step 1 should conform to the
+`ManifoldCoordinate` protocol, which has these requirements:
+
+```swift
+/// A coordinate in a manifold's global coordinate system.
+public protocol ManifoldCoordinate: Differentiable {
+  /// The global coordinate corresponding to `local` in the chart centered around `self`.
+  @differentiable(wrt: local)
+  func global(_ local: LocalCoordinate) -> Self
```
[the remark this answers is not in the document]
dan-zheng: Hi Frank!

Question about exponential map: the first version you shared (`Vector3 -> Rot3`) is an unary function, but I understand exponential map to be a binary function: `(T, T.TangentVector) -> T`.

Could you please clarify what the first version means?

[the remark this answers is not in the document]
marcrasi: I'm not completely sure I understand your comment. But I'm very open to finding a better name for these things.

> I got confused between what was mutating and what is functional.

None of the requirements in this file are mutating. (There is the `move(along:)` requirement in `Differentiable` which is a mutating version of exp.)

> Rot3::Expmap: Vector3 -> Rot3 (functional): v -> exp(v)

By `Rot3 (functional)`, do you mean that you're thinking of the `Rot3` as a function by the group action on itself?

If I've interpreted that correctly, we can't have an `ExpMap: Vector3 -> Rot3 (functional)` for a general `Manifold` because the general `Manifold` doesn't have a group operation.

We could of course have an `ExpMap`-like function on `LieGroup` (if we eventually define a `LieGroup`).

> Rot3::expmap: Rot3 * Vector3 -> Rot3 (still functional, but also composes): R, v -> R*exp(v)

This is the one that corresponds most closely to the current `ManifoldCoordinate.global` in this PR. That's because `ManifoldCoordinate.global` is a method that takes a `self` parameter (the base point for the exp map). e.g. it's signature for `Rot3Coordinate` is `Rot3Coordinate.global: (Rot3Coordinate) -> (Vector3) -> Rot3Coordinate`.

Using "exponential map" in the names here might not be a good idea because this is a general manifold, not a lie group.

Here's another name idea:
* `global` -> `retraction`
* `local` -> `inverseRetraction`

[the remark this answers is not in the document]
dellaert: `Expmap` is a unary function which is around the origin. It is really closest to the mathematical concept: all matrix groups define an exponential map with is really the matrix exponential, and it is a map from Lie algebra to Lie group. Unary.

 `expmap`, lowercase, is really the "right-trivialization" (if I'm not messing up this definition - always a bit mysterious): we compose a rotation on the right with the exponential map.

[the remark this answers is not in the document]
dellaert: @marcrasi you are right: manifolds don't necessarily have a group operation - let me regroup (no pun intended!)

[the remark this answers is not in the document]
dellaert: So, in GTSAM, we call `global` just `retract`, which matches Absil book.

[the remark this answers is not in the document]
dellaert: but I never liked `localCoordinates` all that much. I'm not in love with `inverseRetract` either, though.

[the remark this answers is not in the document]
dellaert: To amend my comment to allow for manifolds: *if* T is a Lie group, both Expmap and expmap exist, we implement `retract` with `expmap`. If not, we do something manifold-specific.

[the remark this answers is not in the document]
marcrasi: `retract` (and `inverseRetract`?) would work for me!

It would be slightly more "swifty" to name them `retracted` and `inverseRetracted` because of the guideline "When the operation is naturally described by a verb, use the verb’s imperative for the mutating method and apply the “ed” or “ing” suffix to name its nonmutating counterpart." (https://swift.org/documentation/api-design-guidelines/).

But I wouldn't object to calling it `retract` for consistency with existing terminology.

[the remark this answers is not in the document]
marcrasi: Other random name ideas: `unretract`, `localized`.

> To amend my comment to allow for manifolds: if T is a Lie group, both Expmap and expmap exist, we implement retract with expmap. If not, we do something manifold-specific.

Automatically implementing `retract` with `expmap` for Lie groups could be done with protocol extensions in Swift like this:

```
protocol LieGroupCoordinates: ManifoldCoordinates {
  func expmap(_ v: LocalCoordinates) -> Self
}

extension LieGroupCoordinates {
  func retract(_ v: LocalCoordinates) -> Self {
    expmap(v)
  }
}
```

marcrasi:
> Migrated Pose2

This commit doesn't take advantage of all the possible simplifications. It should be possible to completely remove the custom derivatives for `Pose2.init`, `Pose2.rot`, `Pose2.t`, as long as you define the correct `global` (aka `retract`) and `local` (aka `localCoordinates` or `inverseRetract`) methods.

dellaert:
I think `retract` and `localCoordinates` are still the most descriptive and they match the book (at least for retract) and GTSAM. I was trying on `local` for size as it is short and sweet, matching `retract`, but it really does not capture the coordinates aspect - and users new to this need all the help they can get.

ProfFan:
@marcrasi Sorry to be late for the party here, but can you do a small review on it? So I can see what did I do wrong :)

dellaert:
So, the experience of trying to implement Rot3 raised a lot of questions for me. The main one: why do we have the CoordinateStorage indirection? This feels quite heavy to me: I want to just implement Rot3 as having a 3x3 matrix R. Why do I need to create *another* type called Matrix3Coordinate? It seems a lot of what happened in Rot3 was passing through stuff to that type: why not have just 1 type?

Now, I suspect it might have to do something with your comment on Pose2 above: maybe the pass-through is all automatic. and @ProfFan and I just didn't *get* it yet :-) I guess there might be two things going on here at the same time, and I don't yet have the Swiftness of disambiguating them: (a) protocols/module signatures, (b) generation of automatic functionality.

dellaert:
Another comment, which *could* impact the discussion. One of the things that is *regrettably* complex in GTSAM is exactly Rot3: there are two code paths: one that stores a Matrix3, and one that stores a Quaternion. You have to select this via a compile flag. Very clumsy.

In fact, there is a mathematical story that is not entirely captured: both 3x3 matrices and quaternions are valid *representations* for SO(3) (although, one could argue quaternions are not, really). So, there is *group* SO(3) which is coordinate free, and you have several different representations to indicate an element of that group. You could argue that one should be able to use both interchangeably.

However - as my previous comment indicates: KISS! I'd lean towards trying to eliminate extraneous  types (like the storage thingy) in favor of as little code overhead as possible to create your own manifold and/or Lie group.

Sources/SwiftFusion/Geometry/Pose2.swift
```diff
@@ -35,32 +35,110 @@ import TensorFlow
...
+  
+  @differentiable public var t: Vector2 { coordinate.t }
+  
+  @differentiable public var rot: Rot2 { coordinate.rot }
+  
+  /// Twist coordinates, with the first component for rotation and the remaining components for
+  /// translation.
+  ///
+  /// See the documentation comment on the declaration of `Pose2` for more information.
+  public typealias TangentVector = Vector3
```
marcrasi, line 67: This should not be necessary. It is automatically inferred to be equal to Pose2Coordinate.LocalCoordinate.

Sources/SwiftFusion/Geometry/Pose2.swift
```diff
@@ -35,32 +35,110 @@ import TensorFlow
...
+    let r = matmul(cholesky(covariance), Tensor<Double>(randomNormal: [3, 1])).scalars
+    let tv = Pose2.TangentVector(r[0], r[1], r[2])
+    self.move(along: tv)
+  }
+}
+
+// MARK: - Global Coordinate System
+
+public struct Pose2Coordinate: Equatable, KeyPathIterable {
+  var tStorage: Vector2
```
marcrasi, line 100: There is no need to have stored properties `tStorage` and `rotStorate` with wrappers `t` and `rot`.

Instead, you can just have stored properties `t: Vector2` and `rot: Rot2`.

Sources/SwiftFusion/Geometry/Pose2.swift
```diff
@@ -35,32 +35,110 @@ import TensorFlow
...
+  public var t: Vector2 { tStorage }
+  
+  @differentiable
+  public var rot: Rot2 { rotStorage }
+  
+  /// Twist coordinates, with the first component for rotation and the remaining components for
+  /// translation.
+  ///
+  /// See the documentation comment on the declaration of `Pose2` for more information.
+  public typealias TangentVector = Vector3
```
marcrasi, line 113: You shouldn't need this also. The `TangentVector` type will be inferred from the types you use in `local` and `global`.

Sources/SwiftFusion/Geometry/Pose2.swift
```diff
@@ -35,32 +35,110 @@ import TensorFlow
...
+  
+  /// Creates a tangent vector given rotation (`w`) and translation (`v`) components.
+  static func tangentVector(w: Vector1, v: Vector2) -> Vector3 {
+    Vector3(w.x, v.x, v.y)
+  }
+}
+
+// MARK: Custom Derivatives
 
+extension Pose2Coordinate {
```
marcrasi, line 136: Now the interesting stuff starts happening. You should be able to simply delete this entire extension, because the custom derivatives for `init`, `t`, and `rot` are no longer needed.

Sources/SwiftFusion/Geometry/Pose2.swift
```diff
@@ -114,106 +192,78 @@ public struct Pose2: Differentiable, TangentStandardBasis, KeyPathIterable {
...
+  /// Inverse of the rotation.
+  @differentiable
+  func inverse() -> Pose2Coordinate {
+    Pose2Coordinate(self.rot.inverse(), self.rot.unrotate(-self.t))
   }
+}
 
-  /// Moves `self` by `exp(hat(direction))`.
+extension Pose2Coordinate: ManifoldCoordinate {
   public mutating func move(along direction: Vector3) {
```
marcrasi, line 220: Delete this `move`. The automatically-derived `move` function for `ManifoldCoordinate` is the correct one for it.

Sources/SwiftFusion/Geometry/Pose2.swift
```diff
@@ -114,106 +192,78 @@ public struct Pose2: Differentiable, TangentStandardBasis, KeyPathIterable {
...
     self.tStorage.move(along: rot.rotate(v))
     self.rotStorage.move(along: w)
   }
-
-  /// Storage for the pose's translation.
+  
+  /// p * Exp(q)
+  @differentiable(wrt: local)
+  public func global(_ local: Vector3) -> Self {
+    self * Pose2Coordinate(Rot2(local.x), Vector2(local.y, local.z))
```
marcrasi, line 229: The "Exp(q)" here should actually be the last formula in section 4.1 of https://github.com/borglab/gtsam/blob/develop/doc/LieGroups.pdf , because that is the formula for the exp on twist coordinates.

Here's an attempt at it, which I haven't compiled or tested, but if you run the tests it will probably reveal any problems:
```
let (w, v) = Pose2Coordinate.decomposed(tangentVector: local)
let vPerpOverW = Vector2(-v.y / w, v.x / w)
let expW = Rot2(w)
let expTwist = Pose2(expW, vPerpOverW - expW.rotate(vPerpOverW))
return self * expTwist
```

(the fact that we have to calculate the derivative of this thing any time we calculate the derivative of a function on Pose2 raises interesting questions about efficiency, but it's probably best to get something nice and simple working before worrying about that...)

Sources/SwiftFusion/Geometry/Pose2.swift
```diff
@@ -114,106 +192,78 @@ public struct Pose2: Differentiable, TangentStandardBasis, KeyPathIterable {
...
-
-  public init(randomWithCovariance covariance: Tensor<Double>) {
-    self.init(0, 0, 0)
-    let r = matmul(cholesky(covariance), Tensor<Double>(randomNormal: [3, 1])).scalars
-    let tv = Pose2.TangentVector(r[0], r[1], r[2])
-    self.move(along: tv)
+  @differentiable(wrt: global)
+  public func local(_ global: Self) -> Vector3 {
+    let d = self.inverse() * global
+    return Vector3(d.rot.theta, d.t.x, d.t.y)
```
marcrasi, line 243: Of course you'll have to change the log here to be the appropriate inverse if you change the exp above.

Sources/SwiftFusion/Geometry/Pose2.swift
marcrasi: Now that I wrote that out, I notice that it does not work well for `w` near `0`, and it completely fails for `w == 0`. So for this approach to work, the implementation will need to be changed to be one that works better for `w` near `0`.

marcrasi:
> why do we have the CoordinateStorage indirection?

The reason for this indirection is that it reduces the number of "primitive functions" that we need to define derivatives for, which makes it easier to explain in the recipe, and also easier for users to implement.

(By "primitive functions", I mean the functions that are used to access fields of the manifold type and that are used to create new instances of the manifold type. Since every function on manifolds is built out of "primitive functions", specifying the derivatives of the "primitive functions" determines the automatically-derived derivatives of all functions on the manifold.)

When there is a single `var coordinate` field, that is the only field that we need to specify a derivative for. It's easier to explain how to specify the derivative for `var coordinate` than to explain how to specify derivatives for all the fields in the manifold.

The alternative of defining custom derivatives for each field in the manifold is also possible -- in fact, it's what `Pose2` does before this PR.

So there's currently a tradeoff:

 * Manifold type with `var coordinate`. It is easier to specify the derivatives, but everything else is more annoying because you have to go through the indirection.
 * Manifold type that directly contains all its fields. It is harder to specify the derivatives.

I think there's no way around this tradeoff _with the current Swift AD APIs_. However, we could add a Swift AD feature that gives you the best of both worlds: a way that you can create a `Differentiable` type with arbitrary fields, and jointly specify the derivatives of all the fields in a single custom derivative function.

marcrasi:
This discussion makes me realize that this PR does a few distinct things:

 1. The `var coordinate` indirection that makes it easier to specify the desired derivatives.
 2. The `ManifoldCoordinate.global` / `ManifoldCoordinate.local` methods from which we attempt to automatically derive derivatives that correspond with local coordinates.
 3. The "recipe" that documents how to use these things.

My current feelings about these are:

 1. It's a tradeoff, as discussed in my previous comment. In the medium term, we should add the Swift AD feature that makes the tradeoff unnecessary. In the immediate term, I don't have any strong opinion about whether we use it or not.
 2. It's a bad idea to automatically derive the derivatives from these. This is because of the inefficiency and the bad behavior when `w == 0`: https://github.com/borglab/SwiftFusion/pull/37#discussion_r418411388 . I should change it so that users specify the derivatives directly.
 3. Having a recipe is good. It should be updated to correspond with the final decisions on (1) and (2).

Sources/SwiftFusion/Geometry/Pose2.swift
ProfFan: Yep there is a near-zero approximation for that :)

Sources/SwiftFusion/Geometry/Pose2.swift
ProfFan: Thanks!

ProfFan:
I updated the implementation to incorporate @marcrasi 's suggestions and adapted to GTSAM conventions :)

>  1. It's a tradeoff, as discussed in my previous comment. In the medium term, we should add the Swift AD feature that makes the tradeoff unnecessary. In the immediate term, I don't have any strong opinion about whether we use it or not.

Agreed!

> 2. It's a bad idea to automatically derive the derivatives from these. This is because of the inefficiency and the bad behavior when w == 0: #37 (comment) . I should change it so that users specify the derivatives directly.
Having a recipe is good. It should be updated to correspond with the final decisions on (1) and (2).

I think it should be possible also to automatically detect numerical stability issues in automatic differentiation? Anyway I think the current Swift AD can handle the control flow in `abs(w)<eps`, as the update seemed to have no issues.

marcrasi:
Cool, the updates to Pose2 look great now! That simple near-zero approximation is nice, and makes me feel a bit better about (2) from my previous comment.

> I think it should be possible also to automatically detect numerical stability issues in automatic differentiation?

Very interesting feature idea. We don't do that now.

Maybe it would even be possible for AD to automatically derive approximations when there is a stability issue.

This is probably a very long term thing that we shouldn't spend too much effort on now. 

> Anyway I think the current Swift AD can handle the control flow in abs(w)<eps, as the update seemed to have no issues.

Yep, most control flow is handled quite stably now.

dan-zheng:
> I think it should be possible also to automatically detect numerical stability issues in automatic differentiation?

There is research on programming languages and numerical stability: ["Automatically Improving Accuracy for Floating Point Expressions", Distinguished Paper PLDI 2015](https://herbie.uwplse.org/pldi15.html)

I haven't thought much about incorporating numerical stability analysis or rewriting into Swift. A simple "finite differences" derivative testing function ([early prototype](https://gist.github.com/dan-zheng/70346b0e4529d8e9561e24ede517fbea)) may be able to catch unstable derivatives for continuous functions.

ProfFan:
So, I adapted @dellaert 's PR and added some tests :) However I still get one ICE from `Rot3.swift`, line 15. I commented that out and everything works beautifully.

ProfFan:
I think the *mathematical* cause is that we should not differentiate Rot matrix members as they are constrained internally, but I have no idea why it fails in code... @marcrasi any ideas?

ProfFan:
@dan-zheng Yeah I am very interested in that part as well... However I think my experience on the Swift compiler probably would not allow me to touch that yet...

marcrasi:
@ProfFan, I figured the ICE (Internal Compiler Error?) out. Use `self.init(coordinate:)` instead of `self.init(coordinateStorage:)`.

I have filed a bug: https://bugs.swift.org/browse/SR-12720 !

ProfFan:
@marcrasi I pushed in the Rot3 with proper invariant checks :) Will tackle the Pose3 tomorrow.

ProfFan:
@marcrasi I think I found another issue with this: in Pose3 I need to initialize a `Rot3` with the Expmap, and the most intuitive way to do it would be:

```swift
extension Rot3 {
  @differentiable
  public static func fromTangent(_ vector: Vector3) -> Self {
    var actual = Rot3()
    actual.move(along: vector)
    return actual
  }
}
```

However this triggers a compiler ICE that says:
```swift
1.	Swift version 5.2-dev (LLVM b3057cffb6, Swift da7410955d)
2.	While running pass #855 SILModuleTransform "Differentiation".
3.	While canonicalizing `differentiable_function` SIL node   %16 = differentiable_function [parameters 0 1] %15 : $@convention(method) (Vector3, @inout Rot3) -> () // users: %18, %17
4.	While ...in SIL function "@AD__$s11SwiftFusion4Rot3V11fromTangentyAcA7Vector3VFZ__vjp_src_0_wrt_0".
 for 'fromTangent(_:)' (at /Users/proffan/Projects/Development/CV/SLAM/SwiftFusion/SwiftFusion/Sources/SwiftFusion/Geometry/Rot3.swift:33:10)
5.	While processing // differentiability witness for Rot3.move(along:)
sil_differentiability_witness private [parameters 0 1] [results 0] @$s11SwiftFusion4Rot3V4move5alongyAA7Vector3V_tF : $@convention(method) (Vector3, @inout Rot3) -> () {
}

 on SIL function "@$s11SwiftFusion4Rot3V4move5alongyAA7Vector3V_tF".
 for 'move(along:)' (at /Users/proffan/Projects/Development/CV/SLAM/SwiftFusion/SwiftFusion/Sources/SwiftFusion/Geometry/Rot3.swift:18:19)
6.	While generating VJP for SIL function "@$s11SwiftFusion4Rot3V4move5alongyAA7Vector3V_tF".
 for 'move(along:)' (at /Users/proffan/Projects/Development/CV/SLAM/SwiftFusion/SwiftFusion/Sources/SwiftFusion/Geometry/Rot3.swift:18:19)
7.	While generating pullback for SIL function "@$s11SwiftFusion4Rot3V4move5alongyAA7Vector3V_tF".
 for 'move(along:)' (at /Users/proffan/Projects/Development/CV/SLAM/SwiftFusion/SwiftFusion/Sources/SwiftFusion/Geometry/Rot3.swift:18:19)
```

I think this is somewhat related to higher-order differentiation, is it?

dan-zheng:
> However this triggers a compiler ICE that says:

The meaningful crash message is actually just above the stack trace that you shared: it should be an assertion failure or something similar. Could you please edit your comment to share that?

Sorry about the compiler crashers! Some of them are known issues, we can prioritize fixing them.

ProfFan:
@dan-zheng Sorry for that haha 🤣  it's 2am now and I am a bit dizzy...

I updated it :)

ProfFan:
@dan-zheng Also pushed the source here, so you can take a look if needed :)

marcrasi:
@ProfFan, I pushed a fix for that latest error. It's the same thing, just getting triggered in a different way.

Both of these are caused by accessing `coordinateStorace`, which is (intentionally) not differentiable. (The `move(along:)` method accesses `coordinateStorage`). Hopefully this will become clearer when the compiler gives you a proper error message about not being able to access `coordinateStorage` in a differentiable function :).

> it's 2am now and I am a bit dizzy...

Take care, don't overwork :)!

ProfFan:
Pushed `Pose3` with tests from GTSAM.

@marcrasi @dellaert Please (p)review :)

Sources/SwiftFusion/Core/Manifold.swift
```diff
@@ -0,0 +1,110 @@
...
+  ///
+  /// Note: This is not differentiable, and therefore clients should use `init(coordinate:)`
+  /// instead.
+  init(coordinateStorage: Coordinate)
+}
+
+extension Manifold {
+  /// The coordinate of `self`.
+  @differentiable
+  public var coordinate: Coordinate { coordinateStorage }
```
dellaert, line 53: This swift syntax I still don't understand, even after looking for it in manual: what does the `{ coordinateStorage }` mean here?

Sources/SwiftFusion/Core/Manifold.swift
```diff
@@ -0,0 +1,110 @@
...
+    //
+    // Let `f: Manifold -> Coordinate` be `f(x) = x.coordinateStorage`.
+    //
+    // `D_x(f)` (the derivative of `f` at `x`) is a linear approximation of how changes in local
+    // coordinates around `x` lead to changes in global coordinates around `x.coordinateStorage`.
+    //
+    // `x.coordinateStorage.global: LocalCoordinate -> Coordinate` defines _exactly_ how local
+    // coordinates at `x` map to global coordinates.
+    //
+    // Therefore, `D_x(f)` is the derivative of `x.coordinateStorage.global`.
```
dellaert, line 72: Yes, but close the loop :-) Why does that mean that `coordinateStorage.global` is the pullback? And (see confusion above) why is it not `coordinate.global` ?

Sources/SwiftFusion/Core/Manifold.swift
ProfFan: This means `coordinate` is a computed property that, when accessed, will return `coordinateStorage`

Sources/SwiftFusion/Core/Manifold.swift
```diff
@@ -0,0 +1,110 @@
...
+/// A point on a differentiable manifold.
+public protocol Manifold: Differentiable {
+  /// The manifold's global coordinate system.
+  associatedtype Coordinate: ManifoldCoordinate
+      where Coordinate.LocalCoordinate == Self.TangentVector
+
+  /// The coordinate of `self`.
+  ///
+  /// Note: This is not differentiable, and therefore clients should use `coordinate` instead.
+  var coordinateStorage: Coordinate { get set }
```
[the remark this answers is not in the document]
ProfFan: I think this is a workaround for the fact that currently public variables cannot have custom derivatives.

Sources/SwiftFusion/Core/Manifold.swift
```diff
@@ -0,0 +1,110 @@
+/// A coordinate in a differentiable manifold's global coordinate system.
+///
+/// Includes a coordinate chart [1] between `Self` and `R^n`, centered at `self`.
+///
+/// [1] https://en.wikipedia.org/wiki/Topological_manifold#Coordinate_charts
+public protocol ManifoldCoordinate: Differentiable {
+  /// The local coordinate system used in the chart.
+  ///
+  /// Isomorphic to `R^n`, where `n` is the dimension of the manifold.
+  associatedtype LocalCoordinate: AdditiveArithmetic & Differentiable & VectorProtocol
```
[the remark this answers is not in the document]
marcrasi: This is important, maybe everything will make sense if I can successfully explain why there is a `ManifoldCoordinate.LocalCoordinate` that is distinct from `ManifoldCoordinate.TangentVector`.

The high level idea behind this PR is:

1. When a user defines a struct, Swift can automatically generate a tangent space for the struct as the direct product of the struct's properties' tangent spaces. This is nice because derivatives involving the struct just work without any boilerplate.
2. For SwiftFusion's geometry, we want a different tangent space. We can achieve this by defining a fully custom `TangentVector` type (for an example, see `Pose2.swift` in the master branch as it is today). This involves an inconvenient amount of hard-to-explain boilerplate.
3. The tangent spaces in (1) and (2) are isomorphic to each other. This suggests a strategy for getting the advantages of (1) with the tangent space of (2): Do all the derivatives in the automatically-generated tangent space from (1), but map the derivatives through the isomorphism to (2) before exposing them to the user.

There are many ways to implement the strategy in (3), and this PR is one way. This PR works like this:

* `ManifoldCoordinate.TangentVector` is generated automatically (1).
* `ManifoldCoordinate.LocalCoordinate` is the user-specified tangent space (2).
* `ManifoldCoordinate.global/local` (aka `retract/localCoordinates`) determine the isomorphism between (1) and (2).
* `Manifold` is a wrapper type with `Manifold.TangentVector == ManifoldCoordinate.LocalCoordinate`, so that clients who use the `Manifold` wrapper type get the user-specified tangent space (2) without having to worry about translating between it and the automatically-generated tangent space (1).

Sources/SwiftFusion/Core/Manifold.swift
```diff
@@ -0,0 +1,110 @@
...
+    where LocalCoordinate.TangentVector == LocalCoordinate
```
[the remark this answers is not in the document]
marcrasi: Good point.

It can't currently be part of `VectorProtocol` because `VectorProtocol` does not imply `Differentiable` and so `VectorProtocol` can't specify constraints related to `Differentiable` concepts. (`VectorProtocol` does not imply `Differentiable` because some user might want to do vector calculations without opting in to all the extra baggage of differentiability.)

I can think of some ways to make this better, so I'll make a note of this and we can consider it when rethinking `VectorProtocol`.

Sources/SwiftFusion/Core/Manifold.swift
```diff
@@ -0,0 +1,110 @@
...
+
+  /// The global coordinate corresponding to `local` in the chart centered around `self`.
+  ///
+  /// Satisfies the following properties:
+  /// - `global(LocalCoordinate.zero) == self`
+  /// - There exists an open set `B` around `LocalCoordinate.zero` such that
+  ///   `local(global(b)) == b` for all `b \in B`.
+  @differentiable(wrt: local)
+  func global(_ local: LocalCoordinate) -> Self
```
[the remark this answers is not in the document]
marcrasi: I'll change later today!

Sources/SwiftFusion/Core/Manifold.swift
```diff
@@ -0,0 +1,110 @@
...
+
+  /// The local coordinate corresponding to `global` in the chart centered around `self`.
+  ///
+  /// Satisfies the following properties:
+  /// - `local(self) == LocalCoordinate.zero`
+  /// - There exists an open set `B` around `self` such that `local(global(b)) == b` for all
+  ///   `b \in B`.
+  @differentiable(wrt: global)
+  func local(_ global: Self) -> LocalCoordinate
```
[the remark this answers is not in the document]
marcrasi: I'll change later today!

Sources/SwiftFusion/Core/Manifold.swift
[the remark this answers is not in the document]
marcrasi: It is a workaround for the fact that protocol requirements cannot have default custom derivatives (https://bugs.swift.org/browse/TF-982).

Sources/SwiftFusion/Core/Manifold.swift
marcrasi: Yeah, a recent swift feature is that single-expression function bodies implicitly return the value of the expression. I could start always using `return` if you think that would make things clearer.

Sources/SwiftFusion/Core/Manifold.swift
marcrasi: I have an idea how to update this whole comment to be clearer by using notation closer to the Swift code and by explicitly saying how we get the pullback expression:

```
    // Explanation of this pullback:
    //
    // Let `f: Manifold -> Coordinate` be `f(x) = x.coordinateStorage`.
    //
    // `differential(at: x, in: f)` is a linear approximation of how changes in local
    // coordinates around `x` lead to changes in global coordinates around `x.coordinateStorage`.
    //
    // `x.coordinateStorage.global: LocalCoordinate -> Coordinate` defines _exactly_ how local
    // coordinates around zero map to global coordinates around `x`.
    //
    // Therefore, `differential(at: x, in: f) = differential(at: zero, in: x.coordinateStorage.global)`.
    //
    // The pullback is the dual map of the differential, so taking duals of both sides gives:
    //   `pullback(at: x, in: f) = pullback(at: zero, in: x.coordinateStorage.global)`.
```

marcrasi:
By the way, I started a forum post on a small improvement to Swift AD motivated by this: https://forums.swift.org/t/stored-property-custom-derivatives/36194?u=marcrasi That change is pretty small and noncontroversial, so I expect it to happen soon.

Sources/SwiftFusion/Core/Manifold.swift
dellaert: I knew about that feature. It would indeed, I think, make it more clear to the novel user here, as this is a double whammy otherwise: computed property and this new feature. For example, I did not understand that this could not be used to set. Adding `return` makes it clear that this is just a gettable property (even though storage is `get set`)

Sources/SwiftFusion/Core/Manifold.swift
```diff
@@ -0,0 +1,144 @@
...
+  func vjpCoordinate() -> (value: Coordinate, pullback: (Coordinate.TangentVector) -> TangentVector) {
+
+    // Explanation of this pullback:
+    //
+    // Let `f: Manifold -> Coordinate` be `f(x) = x.coordinateStorage`.
+    //
+    // `differential(at: x, in: f)` is a linear approximation of how changes in local
+    // coordinates around `x` lead to changes in global coordinates around `x.coordinateStorage`.
+    //
+    // `x.coordinateStorage.retract: LocalCoordinate -> Coordinate` defines _exactly_ how local
```
dellaert, line 85: I think you should say here `TangentVector -> Coordinate` to be consistent with pullback above, and also because I think within Manifold we should really only talk about Coordinate/TangentVector once we 'defined` them.

Sources/SwiftFusion/Geometry/Pose3.swift
```diff
@@ -0,0 +1,191 @@
+import TensorFlow
+
+/// TODO: Should be merged to Vector.swift
+public struct Vector6: Differentiable, VectorProtocol, KeyPathIterable, TangentStandardBasis {
```
dellaert, line 4: Maybe `Twist` ? Definitely not a vector6 if you distinguish a v and w part.

Sources/SwiftFusion/Geometry/Pose3.swift
```diff
@@ -0,0 +1,191 @@
...
+
+public struct Pose3Coordinate: Equatable, KeyPathIterable {
+  var t: Vector3
+  var rot: Rot3
+}
+
+public extension Pose3Coordinate {
+  @differentiable
+  init(_ rot: Rot3, _ t: Vector3) {
+    self.t = t
```
dellaert, line 73: respect the order, @ProfFan ! :-)

Sources/SwiftFusion/Geometry/Pose3.swift
```diff
@@ -0,0 +1,191 @@
...
+  public static func fromTangent(_ vector: Vector6) -> Self {
+    return Pose3(coordinate: Pose3Coordinate(Rot3(), Vector3.zero).retract(vector))
+  }
+  
+}
+
+// MARK: - Global Coordinate System
+
+public struct Pose3Coordinate: Equatable, KeyPathIterable {
+  var t: Vector3
```
dellaert, line 66: respect the order, @ProfFan ! :-)

marcrasi:
Thanks for the comments! I have responded inline and added a commit that addresses them.

I have also removed all the new manifold types (Pose2, Rot3, Pose3) from this PR to keep it small and focused. Those types also exist on other branches, and their implementations on other branches have started diverging, so I think this should make it easier to keep track of things.

marcrasi:
Let's do it! I have resolved the conflict and I'll do a squash n' merge once the CI passes.

Of course, further discussion is still very welcome and I'll keep iterating on this if anyone has more thoughts.

doc/DifferentiableManifoldRecipe.md
```diff
@@ -0,0 +1,180 @@
...
+Note: We want to make the recipe simpler. See
+[issue #45](https://github.com/borglab/SwiftFusion/issues/45).
+
+## Steps
+
+### 1. Choose a global coordinate system.
+
+Create a `struct` whose values represent points on the manifold. Every point on
+the manifold must be representable by at least one `struct` value. Every
+`struct` value must represent at most one point on the manifold.
```
dabrahams, line 22: @marcrasi Doesn't this simply translate to: “Every point on the manifold must be representable by exactly one `struct` value?”  If so, why not say it that way?

dabrahams: [I'm a little concerned about the idea that the values must cover the entire manifold (with no caveats), since obviously our numerical representations are limited.

doc/DifferentiableManifoldRecipe.md
```diff
@@ -0,0 +1,180 @@
...
+### 3. Conform to the ManifoldCoordinate protocol.
+
+The coordinate `struct` you chose in step 1 should conform to the
+`ManifoldCoordinate` protocol, which has these requirements:
+
+```swift
+public protocol ManifoldCoordinate: Differentiable {
+  /// The local coordinate type of the manifold.
+  ///
+  /// This is the `TangentVector` of the `Manifold` wrapper type.
```
dabrahams, line 59: What “`Manifold` wrapper type?”  I haven't seen one yet.

doc/DifferentiableManifoldRecipe.md
```diff
@@ -0,0 +1,180 @@
...
+the manifold's tangent vectors during differentiation.
+
+Note: These are also known as the "exponential map" and "logarithm" in
+differential geometry.
+
+### Step 4. Create the manifold type.
+
+Create a struct with a stored property and initalizer as follows:
+```
+var coordinateStorage: ${your coordinate type}
```
dabrahams, line 97: Don't think you should be using gyb syntax here 😉 

It's unfortunate that we don't have a way to embed typographic styles into code snippets in markdown, or we could use italics.  As it stands, I would just put "where `X` is your coordinate type” in the prose, and use “`X`”.

doc/DifferentiableManifoldRecipe.md
```diff
@@ -0,0 +1,180 @@
...
+init(coordinateStorage: ${your coordinate type} {
+  self.coordinateStorage = coordinateStorage
+}
+```
+
+Conform the struct to `Manifold` and specify two typealiases:
+```
+public typealias Coordinate = {your coordinate type}
+public typealias TangentVector = {your local coordinate type}
+```
```
dabrahams, line 108: This whole thing would be better as a copy/paste code template rather than fragments from the inside of some declaration not actually shown.

PR #43, opened 2020-05-08
TF-1234 workaround

marcrasi:


dan-zheng:
@marcrasi @ProfFan 
I'll look into a fix for TF-1234 tonight. Letting you know so we don't do duplicate work!

marcrasi:
I kinda suspect that the TF-1234 fix didn't get merged into `tensorflow` properly, because the TF-1234 reproducer doesn't crash on `master` but does crash on latest `tensorflow`. That's as far as I got in my investigation, and I'm not looking any more tonight.

dan-zheng:
I took a look at the compilation errors, and I think they're actually caused by wrong index subset calculation (related to [TF-874](https://bugs.swift.org/browse/TF-874)). I've encountered this issue before and want to continue investigating tomorrow!

I think the current [TF-1234](https://bugs.swift.org/browse/TF-1234) reproducer does not exactly cover all of the code that's causing compilation errors. I'll look into minimal reproducers for the not-currently-covered issues.

Issue #42, opened 2020-05-07
Main issue for performance investigations


marcrasi:
I'll post information about performance investigations here.

marcrasi:
I ran Pose2SLAMG2O on the INTEL dataset from https://lucacarlone.mit.edu/datasets/.

Here's a flamegraph: https://storage.googleapis.com/swift-tensorflow-misc-files/flamegraphs/2020-05-07-SwiftFusion-Pose2G2O.svg

# Discussion

It's spending a lot of time doing small matrix operations in TensorFlow. The TensorFlow overhead is large relative to the small operations, so most of the time is probably spent on TensorFlow overhead. Therefore, switching from `Tensor` to `Array` (or eventually [`SIMD`](https://developer.apple.com/documentation/swift/simd)) is going to be a huge performance win and we should do that before any other attempts at optimization.

# How to make the flame graph (Linux)

Prerequisites:
* `linux-perf` debian package (or equivalent)
* clone https://github.com/brendangregg/FlameGraph somewhere

```
# start in SwiftFusion repo root
# assumes FlameGraph is cloned to ../FlameGraph
swift build -c release
perf record -F 499 -g -- .build/release/Pose2SLAMG2O ~/input_INTEL_g2o.g2o
perf script | ../FlameGraph/stackcollapse-perf.pl | swift-demangle | ../FlameGraph/flamegraph.pl --width 1800 > flamegraph.svg
```

dellaert:
Am I right in concluding `transpose` is a lot of the CPU? Not used to reading the flamegraphs

marcrasi:
> Am I right in concluding transpose is a lot of the CPU? Not used to reading the flamegraphs

Yes, that's correct! And there's an easy optimization to completely eliminate that call. `matmul`, which is [called from this method](https://github.com/borglab/SwiftFusion/blob/59319421bfd84eb2efa0dfb55b493c79eaf4123d/Sources/SwiftFusion/Inference/JacobianFactor.swift#L88), has `transposed` arguments that let you replace
```
matmul(jacobians[pos].transposed(), r)
```
with
```
matmul(jacobians[pos], transposed: true, r)
```

ProfFan:
Thanks Marc for the benchmarks :) I think judging from the flamegraph the TF engine seems to be doing a lot of work for the `transpose` call? I guess it's a elementwise copy (non-SIMD), but still curious on what exactly is happening under the call.

marcrasi:
[...]

PR #44, opened 2020-05-08
NLFG + Manifold Demo

ProfFan:
This is my attempt on combining #37 and #34 . Thanks to @marcrasi and @dan-zheng (kudos!), all tests are running good with the generic `PriorFactor` and `BetweenFactor`.

Next steps:
- [x] Test Pose3SLAM
- [ ] ... other stuff, please comment :)


dellaert:
Very cool!

ProfFan:
So now the Pose3SLAM test is working, but it is indeed extremely slow (Not only in time, but also in iterations). Need to debug what is happening:
```
error = 0.005291152028057079
error = 0.005291150688696615
error = 0.0052911506878240335
error = 0.005291150687709195
error = 0.00529115068769014
error = 0.0052911506876862505
error = 0.005291150687684503
error = 0.0052911506876833925
error = 0.005291150687682579
error = 0.005291150687681873
error = 0.005291150687681252
error = 0.005291150687680697
error = 0.005291150687680208
error = 0.005291150687679765
error = 0.005291150687679377
```

ProfFan:
Could be a bug of CGLS, need also check #42

ProfFan:
@marcrasi  I think I found the problem: `jacobian()` for `Vector6` is wrong, its first part is empty (always zero). Is this related to the fact that Vector6 is a combination of two `Vector3`s?

ProfFan:
Test:
```swift
  func testPose3SimplePriorFactor() {
    let P = Vector3(3.5,-8.2,4.2)
    let R = Rot3.fromTangent(Vector3(0.3, 0, 0))
    let t12 = Vector6(Tensor<Double>(repeating: 0.1, shape: [6]))
    let t1 = Pose3(R, P)
    let t2 = Pose3(coordinate: t1.coordinate.global(t12))
    let prior_factor = PriorFactor(0, t1)
    
    var vals = Values()
    vals.insert(0, AnyDifferentiable(t1)) // should be identity matrix,
    // but is not (upper left is zero)
    // Change this to t2, still zero in upper left block
    
    print(prior_factor.linearize(vals))
  }
```

Sources/SwiftFusion/Inference/PriorFactor.swift
```diff
@@ -0,0 +1,64 @@
...
+  public func errorVector(_ values: Values) -> T.Coordinate.LocalCoordinate {
+    let val = values[keys[0]].baseAs(T.self)
+    let error = difference.local(val)
+    return error
+  }
+  
+  public func linearize(_ values: Values) -> JacobianFactor {
+    let j = jacobian(of: self.errorVector, at: values)
+    let j1 = Tensor<Double>(stacking: (0..<j.count).map { i in
+      (j[i]._values[0].base as! T.TangentVector).tensor.flattened()
```
marcrasi, line 59: Should this be `(j[i]._values[values._indices[key1]!].base as! T.TangentVector).tensor.flattened()`, like it is in `BetweenFactor`?

I found this while trying to figure out why you are getting the wrong jacobian, though I don't think this is the reason.

I have a change coming soon that will make `linearize` much simpler so that it doesn't need such a complicated confusing expression in it...

ProfFan: Yep I printed out the "raw" jacobian as well and the result is the same. I'll wait for your change then, thanks in advance! :)

marcrasi:
It looks like an AD bug. Here's a workaround: https://github.com/borglab/SwiftFusion/pull/47

ProfFan:
@marcrasi Thanks!

ProfFan:
image

Can confirm Pose3SLAM example is working as expected :) Kudos to @marcrasi and @dan-zheng  for the help!

marcrasi:
> Can confirm Pose3SLAM example is working as expected :) Kudos to @marcrasi and @dan-zheng for the help!

That's awesome! 🎉

ProfFan:
Just a side note: adding @differentiable constraints to the `init(scalars: [Double])` triggers this ICE:
```
Assertion failed: (newAdjointValue.getType() == getRemappedTangentType(originalValue->getType())), function addAdjointValue, file /Users/danielzheng/swift-debug/swift/lib/SILOptimizer/Differentiation/PullbackEmitter.cpp, line 244.
```

marcrasi:
Yeah, I just ran into that. I think it's the same one that you have been running into every time you do some differentiable stuff with structs that have custom tangent vectors (https://bugs.swift.org/browse/TF-969). I'll try to prioritize fixing that soon, so that it doesn't keep happening :)

In the meantime, I know how to work around it, so I think I'll have a fix pretty soon.

dan-zheng:
> ```
> Assertion failed: (newAdjointValue.getType() == getRemappedTangentType(originalValue->getType())), function addAdjointValue, file /Users/danielzheng/swift-debug/swift/lib/SILOptimizer/Differentiation/PullbackEmitter.cpp, line 244.
> ```

I'm happy to take a look at this now, since I've encountered similar errors in the past.
@marcrasi: could you please 👍 if this is cool with you?

marcrasi:
@ProfFan, I have pushed a commit to this PR that fixes it. It was basically adding some `@differentiable` annotations plus working around some of the known AD issues.

marcrasi:
Oh, it seems like a test is still broken. (It didn't run for me locally on Linux because Linux requires registration in XCTestManifests and some of the tests aren't registered there).

I'll investigate the test and add it to XCTestManifests.

marcrasi:
> (It didn't run for me locally on Linux because Linux requires registration in XCTestManifests and some of the tests aren't registered there).

oooo, I can get automatic test discovery on Linux with `swift test --enable-test-discovery` (https://oleb.net/2020/swift-test-discovery/)! I'll use that instead of manually registering all the tests! I'll also eventually send a cleanup PR cleaning up all the manual test registrations.

ProfFan:
@marcrasi Oh I didn't know that! If that is the case I think yeah, let's nuke these :)

marcrasi:
Ok, just pushed a fix for the test. All the tests pass for me on Linux so I'm quite hopeful that the macos CI will also pass.

ProfFan:
@marcrasi I migrated the Rot3 impl to the new Matrix implementation, but it seems we are missing some derivative workarounds for Matrix. I pushed the WIP to feature/new_matrix_for_pose3 , if you have time can you take a look? Thanks in advance!

marcrasi:
> @marcrasi I migrated the Rot3 impl to the new Matrix implementation, but it seems we are missing some derivative workarounds for Matrix. I pushed the WIP to feature/new_matrix_for_pose3 , if you have time can you take a look? Thanks in advance!

Yep, will do! It might be a few hours because I have some meeting soon.

Note: Because Rot3 uses a fixed-size matrix, a fixed sized Matrix3x3 struct would be much more efficient than the dynamically-sized Matrix struct. (This is because the array in the Matrix requires a heap allocation. This is related to the heap allocation stuff we talked about in the meeting on Friday. I have some ideas about this that I will write up soon.). So we should probable generate some fixed-size matrix structs eventually. But for now, Matrix is a good intermediate step.

marcrasi:
> @marcrasi I migrated the Rot3 impl to the new Matrix implementation, but it seems we are missing some derivative workarounds for Matrix. I pushed the WIP to feature/new_matrix_for_pose3 , if you have time can you take a look? Thanks in advance!

I took a look and discovered that it would require tons of boilerplate right now, boilerplate that would be reduced by other work (https://bugs.swift.org/browse/TF-982 + the `DifferentiableVector` mentioned here: https://github.com/borglab/SwiftFusion/pull/37#discussion_r421730219). Therefore, I propose that we leave the Rot3 matrix operations as Tensor for now and revisit them when some of that other work is done. Since the Tensor doesn't leak out of Rot3's API, the only consequence is that Rot3 will be a bit slow until w get around to fixing it. Other parts of SwiftFusion should be unaffected.

marcrasi:
What do you think about just squash n' merging this into master and then addressing any comments as additional small PRs into master? We're all basically already doing all our work off this branch, so it would be convenient to have it in master so all our work can be small PRs into master.

Let me know if you'd like any help resolving the conflicts and fixing up calls to changed APIs.

ProfFan:
@marcrasi Yeah that's cool with me! I can resolve the conflicts and squash, then we can move on from this and close the other stale PRs. @marcrasi @dan-zheng @dellaert Please comment :)

ProfFan:
@marcrasi Done :)

PR #48, opened 2020-05-11
replace Tensor with array-backed Vector and Matrix

marcrasi:
This PR adds array-backed `Vector` and `Matrix` types, and replaces most uses of `Tensor` with them.

I renamed the existing `Vector.swift.gyb` to `VectorN.swift.gyb`, so that I could free up the name `Vector.swift` for the dynamically-sized `Vector` type.

This gives me a 15x speedup on CGLS: before, 200 iterations of CGLS on INTEL took 90s; after, they take 6s. This is not the full 900x speedup that I mentioned in https://github.com/borglab/SwiftFusion/issues/42#issuecomment-626142882 because the full speedup involves more changes and I wanted to keep this PR as small as possible.

### DIFF
--- a/Examples/Pose2SLAMG2O/main.swift
+++ b/Examples/Pose2SLAMG2O/main.swift
@@ -66,15 +66,15 @@ func main() {
   print("Initial error: \(problem.graph.error(val))")
   for _ in 0..<10 {
     let gfg = problem.graph.linearize(val)
-    let optimizer = CGLS(precision: 1e-6, max_iteration: 20)
+    let optimizer = CGLS(precision: 1e-6, max_iteration: 200)
     var dx = VectorValues()
     for i in 0..<val.count {
-      dx.insert(i, Tensor<Double>(shape: [3, 1], scalars: [0, 0, 0]))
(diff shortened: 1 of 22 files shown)


Examples/Pose2SLAMG2O/main.swift
```diff
@@ -66,15 +66,15 @@ func main() {
   print("Initial error: \(problem.graph.error(val))")
   for _ in 0..<10 {
     let gfg = problem.graph.linearize(val)
-    let optimizer = CGLS(precision: 1e-6, max_iteration: 20)
+    let optimizer = CGLS(precision: 1e-6, max_iteration: 200)
```
ProfFan, line 69: Is `20` not converging here?

marcrasi: Yeah, when I run it on the INTEL dataset it seems to take 400ish iterations for the CGLS to converge to the `1e-6` precision. I previously put it at 20 because it was so slow and I wanted to see everything working without waiting forever.

ProfFan: I see ;)

PR #49, opened 2020-05-11
Make Values.TangentVector == VectorValues

marcrasi:
This sets `Values.TangentVector` to `VectorValues`, which allows a few nice improvements to the APIs:
* To update a `Value` using a `VectorValues`, now all you need is `vals.move(along: vectorval)` instead of needing to loop over all the components and cast them.
* We can now write a generic `JacobianFactor` initializer that linearizes any error function, which simplifies the implementation of `linearize`.

Other API changes:
* Now you access values using `vals[key, as: Pose2.self]` instead of `vals[key].baseAs(Pose2.self)`. This is necessary so that `Values` knows what type you are accessing and therefore knows how to convert its tangent vector to a `Vector` suitable for inclusion in `VectorValues`.
* Now you insert values with `vals.insert(key, value)` instead of `vals.insert(key, AnyDifferentiable(value)`. (It still gets stored as a type-erased `AnyDifferentiable` though).

PR #50, opened 2020-05-11
Make Values.TangentVector == VectorValues

marcrasi:
(This is https://github.com/borglab/SwiftFusion/pull/49, rebased to feature/nlfg_with_manifold. I'm pasting the same description below.)

This sets `Values.TangentVector` to `VectorValues`, which allows a few nice improvements to the APIs:
* To update a `Value` using a `VectorValues`, now all you need is `vals.move(along: vectorval)` instead of needing to loop over all the components and cast them.
* We can now write a generic `JacobianFactor` initializer that linearizes any error function, which simplifies the implementation of `linearize`.

Other API changes:
* Now you access values using `vals[key, as: Pose2.self]` instead of `vals[key].baseAs(Pose2.self)`. This is necessary so that `Values` knows what type you are accessing and therefore knows how to convert its tangent vector to a `Vector` suitable for inclusion in `VectorValues`.
* Now you insert values with `vals.insert(key, value)` instead of `vals.insert(key, AnyDifferentiable(value)`. (It still gets stored as a type-erased `AnyDifferentiable` though).

### DIFF
--- a/Examples/Pose2SLAMG2O/main.swift
+++ b/Examples/Pose2SLAMG2O/main.swift
@@ -36,7 +36,7 @@ struct G2OFactorGraph: G2OReader {
   var graph: NonlinearFactorGraph = NonlinearFactorGraph()
 
   public mutating func addInitialGuess(index: Int, pose: Pose2) {
-    initialGuess.insert(index, AnyDifferentiable(pose))
+    initialGuess.insert(index, pose)
   }
 
   public mutating func addMeasurement(frameIndex: Int, measuredIndex: Int, pose: Pose2) {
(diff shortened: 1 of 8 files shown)


marcrasi:
Hmm, this build failure is very mysterious because it doesn't happen on Linux and also this PR doesn't seem to touch anything related to it. I'll investigate.

marcrasi:
I figured out the compile error. I have included a workaround, and I also made a proper fix that should appear in toolchains eventually: https://github.com/apple/swift/pull/31723

ProfFan:
@marcrasi This PR is great :) Since the `feature/nlfg_with_manifold` branch is merged into master, should we change the base of this PR to `master` as well?

marcrasi:
> @marcrasi This PR is great :) Since the feature/nlfg_with_manifold branch is merged into master, should we change the base of this PR to master as well?

Oh yes, done.

PR #53, opened 2020-05-12
genericize CGLS

marcrasi:
This PR introduces a `DecomposedAffineFunction` protocol abstracting the concept of an affine function whose linear and bias components are accessible. Also, its linear component can be applied in the forwards or adjoint direction.

 This PR also changes CGLS to work in terms of an abstract `DecomposedAffineFunction`. Finally, this PR conforms `GaussianFactorGraph` to `DecomposedAffineFunction` so that we can run CGLS on it.

This allows us to run CGLS on anything, as long as it supports the required operations. For example:
* `GaussianFactorGraph` as it is today.
* We can experiment with other representations of gaussian factor graphs that involve fewer heap allocations, without modifying the implementation of CGLS, and without messing with the existing `GaussianFactorGraph`!
* Matrix free jacobians.
* In principle, it should be possible to write a `DistributedGaussianFactorGraph` that distributes its state over multiple computers, and run the existing CGLS algorithm without any modifications. (I haven't thought too hard about this, so there may be some practical obstacles.)

I wrote more justifications for why genericizing is good at https://github.com/borglab/SwiftFusion/issues/42#issuecomment-627499614 and https://github.com/borglab/SwiftFusion/issues/42#issuecomment-627576968.


Examples/Pose2SLAMG2O/main.swift
```diff
@@ -71,7 +71,7 @@ func main() {
     for i in 0..<val.count {
       dx.insert(i, Vector(zeros: 3))
     }
-    optimizer.optimize(gfg: gfg, initial: &dx)
+    optimizer.optimize(objective: gfg, initial: &dx)
```
ProfFan, line 74: This line looks much more natural now :)

Sources/SwiftFusion/Inference/Errors.swift
```diff
@@ -18,43 +18,36 @@ import TensorFlow
...
-  public var norm: Double {
+  public var squaredNorm: Double {
     get {
-      self.map { $0.squared().sum() }.reduce(0.0, { $0 + $1 })
+      values.base.map { $0.squared().sum() }.reduce(0.0, { $0 + $1 })
     }
   }
-  
-  /// Errors + scalar
-  static func + (_ lhs: Self, _ rhs: Double) -> Self {
```
ProfFan, line 41: Could you explain briefly how these are automatically implemented? Curious on the workings...

Sources/SwiftFusion/Optimizers/CGLS.swift
```diff
@@ -32,30 +32,31 @@ public class CGLS {
   /// Optimize the Gaussian Factor Graph with a initial estimate
   /// Reference: Bjorck96book_numerical-methods-for-least-squares-problems
   /// Page 289, Algorithm 7.4.1
-  public func optimize(gfg: GaussianFactorGraph, initial: inout VectorValues) {
+  public func optimize<Objective: MatrixLinearLeastSquaresObjective>(
```
ProfFan, line 35: So objective is inferred automatically? That’s nice!

dellaert:
See my comment in #42 - might impact above

Sources/SwiftFusion/Inference/Errors.swift
marcrasi: Yeah! We have automatic synthesis for `AdditiveArithmetic` and `VectorProtocol` that looks at all the fields of the struct. If all the fields conform to `AdditiveArithmetic` or `VectorProtocol`, it automatically implements the requirements by applying the functions to all the members.

marcrasi: The single field `Array<Error>.DifferentiableView` conforms to `AdditveArithmetic` and `VectorProtocol` because:
1. `Error` (aka `Vector`) conforms to them.
2. `Array.DifferentiableView` conforms to `AdditiveArithmetic` when its `Element` does: https://github.com/apple/swift/blob/4bc72aedb2b32c33ed8e2ec241615fc890b60002/stdlib/public/Differentiation/ArrayDifferentiation.swift#L98
3. `Array.DifferentiableView` conforms to `VectorProtocol` when its `Element` does: https://github.com/tensorflow/swift-apis/blob/a8a24c46e478ce50c1c9a7718a41eec453e5b670/Sources/TensorFlow/StdlibExtensions.swift#L224

Sources/SwiftFusion/Optimizers/CGLS.swift
marcrasi: yep

marcrasi:
Based on discussions in #42 and with Dave, I have improved the protocol naming.

dellaert:
We really need to talk :-) I sent an invite for a hangout at 11am your time.

Sources/SwiftFusion/Inference/VectorValues.swift
```diff
@@ -110,6 +99,53 @@ extension VectorValues: AdditiveArithmetic {
...
+
+  public func scaled(by scalar: Double) -> VectorValues {
+    var result = self
+    result.scale(by: scalar)
+    return result
+  }
+
+  // MARK: - Additional EuclideanVectorSpace requirements.
+
+  public var squaredNorm: Double {
```
ProfFan, line 145: Do we still need the `.differentiableMap` methods?

marcrasi: `.differentiableMap` is still necessary for writing a differentiable function that does a map over an array.

This function currently isn't declared `@differentiable`, so it does not need to use `.differentiableMap`.

I like to avoid adding the `@differentiable` annotation until we actually need to use a function in a differentiable way, to keep things as simple as possible.

PR #57, opened 2020-05-15
"protocol FactorGraph" and a Metropolis algorithm

marcrasi:
@dellaert suggested that I try implementing Metropolis to get a feel for factor graph algorithms other than iterative linearization for least squares, so I have done this here. In the process, I have made an initial attempt at a _protocol oriented_ factor graph abstraction.

`protocol FactorGraph` (attempts to) abstractly describe what a factor graph is, leaving the concrete implementation as unconstrained as possible. I depend on [Penguin Graphs](https://saeta.github.io/penguin/graphs/index.html) (being actively developed by Brennan and others) for the graph aspect of the abstraction.

`FactorGraph.metropolisStep` implements a Metropolis step. It works on anything that conforms to `FactorGraph`, and that also satisfies a few additional requirements that are required for the Metropolis algorithm.

`PoseSLAMFactorGraph` is a highly concrete struct that conforms to `FactorGraph`. It's not a general-purpose factor graph. It's there to test out and demonstrate the abstractions.

What do others think about this approach?

Next steps along these lines could be:
* It's pretty straightforward to make the general-purpose `NonlinearFactorGraph` and `JacobianFactorGraph` conform to these protocols. Once we do that, we can run Metropolis on them too.
* Make some new general-purpose factor graph structs that have better performance characteristics than the current `NonlinearFactorGraph` and `JacobianFactorGraph`.
* It should be pretty straightforward to replace the existing iterative linearization & CGLS with a generic algorithm that operates on arbitrary `FactorGraph`s. Once we do that, we can run iterative linearization & CGLS on arbitrary `FactorGraph`s.

I'm pretty interested in trying to do any or all of these, but I'll wait until we have a discussion about the approach before actually starting any :)

### DIFF
--- a/Sources/SwiftFusion/FactorGraph/FactorGraph.swift
+++ b/Sources/SwiftFusion/FactorGraph/FactorGraph.swift
@@ -1,0 +1,72 @@
+// Copyright 2020 The SwiftFusion Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import PenguinGraphs
+
+/// A bipartite graph of "factor vertices" and "variable vertices", where each factor vertex is an
+/// error function over the variables that it is connected to.
+public protocol FactorGraph {
+
+  // MARK: - Graph.
+
+  /// The type of the graph.
+  associatedtype Graph: GraphProtocol
+
+  /// Bipartite graph of "factor vertices" and "variable vertices".
+  var graph: Graph { get }
+
+  // MARK: - Variables.
+
+  /// The type representing the values of the variables.
+  associatedtype Variables: IndexedVariables
+
+  /// Returns `variable`'s vertex id in the graph.
+  ///
+  /// Note: This is a workaround for the fact that there is nothing like a BipartiteGraph protocol.
+  /// If there were, we could simply set Varaibles.Index == Graph.RightVertexId and not have to
+  /// convert between them.
+  func vertexId(of variable: Variables.Index) -> Graph.VertexId
+
+  /// Returns the variable index for the variable vertex `vertex`.
+  ///
+  /// Precondition: `vertex` refers to a variable vertex (not a facor vertex).
+  ///
+  /// Note: This is a workaround for the fact that there is nothing like a BipartiteGraph protocol.
+  /// If there were, we could simply set Varaibles.Index == Graph.RightVertexId and not have to
+  /// convert between them.
+  func variableIndex(of vertex: Graph.VertexId) -> Variables.Index
+
+  // MARK: - Factor error function.
+
+  /// Returns the error of the factor `factor`, at `point`.
+  ///
+  /// This can be interpreted as an unnormalized negative log probability.
+  ///
+  /// Precondition: `factor` refers to a factor vertex.
+  func error(of factor: Graph.VertexId, at point: Variables) -> Double
+}
+
+/// Graph traversal helpers.
+extension FactorGraph where Graph: IncidenceGraph {
+  /// Returns the vertex ids of the factors neighboring `variableIndex`.
+  public func neighborFactors(of variableIndex: Variables.Index) -> [Graph.VertexId] {
+    let variableVertexId = self.vertexId(of: variableIndex)
+    return graph.edges(from: variableVertexId).map { graph.destination(of: $0) }
+  }
+
+  /// Returns the variable ids of the variables neighboring `factor`.
+  public func neighborVariables(of factor: Graph.VertexId) -> [Variables.Index] {
+    return graph.edges(from: factor).map { self.variableIndex(of: graph.destination(of: $0)) }
+  }
+}
--- a/Sources/SwiftFusion/Inference/Metropolis.swift
+++ b/Sources/SwiftFusion/Inference/Metropolis.swift
@@ -1,0 +1,83 @@
+// Copyright 2020 The SwiftFusion Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+import PenguinGraphs
+
+/// Metropolis algorithm.
+extension FactorGraph where
+  // The Metropolis algorithm needs to query the graph for neighboring vertices:
+  Graph: IncidenceGraph,
+  // The Metropolis algorithm needs to see what the variables look like after they have been
+  // updated, without necessarily writing to the variables' underlying storage:
+  Variables: UpdatedViewable,
+  // The Metropolis algorithm needs to know which variables are affected by an update.
+  Variables.Update: IndexedVariables, Variables.Index == Variables.Update.Index
+{
+
+  /// Returns a draw from a proposal distribution at the given starting state.
+  ///
+  /// - Returns:
+  ///   - update: Describes the change from the starting state to the proposed next state.
+  ///   - logQRatio: `log(Q(nextState | startState) / Q(startState | nextState))`, where `Q`
+  ///                is the probability density of the proposal distribution.
+  public typealias MetropolisProposal<Generator: RandomNumberGenerator> =
+    (Variables, inout Generator) -> (update: Variables.Update, logQRatio: Double)
+
+  /// Attempts a Metropolis step on `state` using the given `proposal` distribution, and then
+  /// returns and `Update` describing the step to the new state on acceptance, or returns `nil` on
+  /// rejection.
+  ///
+  /// This algorithm exploits the factor graph structure to avoid recomputing factor errors of
+  /// unaffected factors.
+  ///
+  /// - Parameters:
+  ///   - applyUpdate: Returns the given `Variables`, with the given `Update` applied.
+  public func metropolisStep<Generator: RandomNumberGenerator>(
+    startState: Variables,
+    proposal: MetropolisProposal<Generator>,
+    using generator: inout Generator
+  ) -> Variables.Update? {
+    // Draw from the proposal distribution.
+    let (update, logQRatio) = proposal(startState, &generator)
+    let newState = startState.updatedView(update)
+
+    // Only factors neighboring variables that have changed in the new state could have changed.
+    let possiblyChangedFactorIds: [Graph.VertexId] = update.indices.flatMap(neighborFactors)
+
+    // Compute the acceptance ratio.
+    let logA = possiblyChangedFactorIds.map { possiblyChangedFactorId in
+      return error(of: possiblyChangedFactorId, at: startState)
+        - error(of: possiblyChangedFactorId, at: newState)
+    }.reduce(logQRatio, +)
+
+    // Decide whether we should accept. If not, exit.
+    guard logA >= 0 || Double.random(in: 0...1, using: &generator) < exp(logA) else {
+      // We reject the update.
+      return nil
+    }
+
+    // Update the state with the proposal.
+    return update
+  }
+}
+
+/// A type that can be viewed as having an update applied to it.
+public protocol UpdatedViewable {
+  /// Specifies an update.
+  associatedtype Update
+
+  /// Returns a view of `self` with `update` applied.
+  func updatedView(_ update: Update) -> Self
+}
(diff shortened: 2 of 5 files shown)


dellaert:
This is great, and will drive innovation!

I'll hold off on a detailed review because I think your current FactorGraph protocol is too much Graph, and too little *BipartiteGraph*. Maybe a concept missing from PenguinGraph? BTW the PenguinGraph protocol we actually use seems very light. Do we really need to import the whole thing? Relying on other experimental code is against VC advice :-)

I think FactorGraph = "Bipartite graph where type1 = variables and type2 = factors".

Maybe it's all there and I don't see it, but I would expect the FactorGraph protocol to be symmetric with respect to variables and factors, even though we should allow for implementations that are not.

That brings up interesting points, that we already discussed: some (most!) algorithms will iterate over factors and need fast lookup of variables given the factors. Others, like Gibbs Sampling, will need to pick a variable and need to find first associated factors, then associated variables to discover its Markov blanket. So, perhaps we need protocols that specialize FactorGraph to bring those asymmetric capabilities.

marcrasi:
> I think your current FactorGraph protocol is too much Graph, and too little BipartiteGraph. Maybe a concept missing from PenguinGraph?

> I think FactorGraph = "Bipartite graph where type1 = variables and type2 = factors".

I agree. I think that the ideal is:

```
protocol BipartiteGraph {
  associatedtype LeftVertexId
  associatedtype RightVertexId
}

protocol FactorGraph {
  // - MARK: Graph.
  associatedtype Graph: BipartiteGraph
  var graph: Graph { get }

  typealias VariableId = BipartiteGraph.LeftVertexId
  typealias FactorId = BipartiteGraph.RightVertexId

  // MARK: - Variables.

  associatedtype Variables: IndexedVariables where Variables.Index == VariableId

  // MARK: - Factors.

  func error(of factor: FactorId, at point: Variables) -> Double
}
```

It would be nice if `BipartiteGraph` existed in PenguinGraphs so that we could use it. But we shouldn't let ourselves get stuck waiting -- we can just make a `BipartiteGraph` ourselves right now.

> BTW the PenguinGraph protocol we actually use seems very light. Do we really need to import the whole thing? Relying on other experimental code is against VC advice :-)

If the dependency introduces more friction and churn than it's worth, then we shouldn't import it. My guess is that it won't introduce much friction and churn because it's being developed by others on my team, and SwiftFusion is currently a big priority for my team, and therefore SwiftFusion's needs will heavily influence PenguinGraph's direction.

We can discuss this point at the meeting today.

> Maybe it's all there and I don't see it, but I would expect the FactorGraph protocol to be symmetric with respect to variables and factors, even though we should allow for implementations that are not.

Not sure I get what you're saying here. The FactorGraph protocol that I defined currently has a fundamental asymmetry:
* There is a concept of a value of the variables. (This is the `associatedtype Variables: IndexedVariables`).
* There is a concept of evaluating a factor at a certain variable value. (This is the `func error(of factor: Graph.VertexId, at point: Variables) -> Double`)
Does that asymmetry make sense to you? Do you think it should be more symmetric?

The hypothetical `protocol BipartiteGraph` is fully symmetric though.

> That brings up interesting points, that we already discussed: some (most!) algorithms will iterate over factors and need fast lookup of variables given the factors. Others, like Gibbs Sampling, will need to pick a variable and need to find first associated factors, then associated variables to discover its Markov blanket. So, perhaps we need protocols that specialize FactorGraph to bring those asymmetric capabilities.

Yes!!!! This is exactly the protocol oriented way of thinking.

My favorite way of discovering these specialized protocols is to incrementally do two things:
* implement algorithms, and add protocols describing the capabilities they need
* implement concrete factor graph data structures conforming to the protocols

When we implement algorithms, there is a natural tendency to add too many capabilities to the protocol because we don't realize that a capability we want can be expressed in terms of a more general capability. This tendency is balanced out when we implement concrete factor graph data structures, get annoyed about how many capabilities we need to explicitly implement, and go find ways to delete capabilities from the protocols without breaking the algorithms.

dellaert:
I think we are starting to see eye to eye more and more ;-)

Why does a Graph have indices, though? Can't I just have variables and factors? Do they *need* indices? Seems indices is an implementation detail to store the graph structure. In CS: a graph is (V, E), and a bipartite graph is (V, F, E). No indices required.

Also, your error function has access to all variables, but I think that's too much visibility: I just need my local variable set. 

So, a factor graph has additional semantics on top of (V,F,E): if a factor has a local error function , and that function has a given arity (a binary factor, say), then there better be two edges. *and* the type of the variables connected on the other end has to match up. Not sure how to say all this with protocols.

Finally, a Swift/Protocol question. Why

```
protocol FactorGraph {
  associatedtype Graph: BipartiteGraph
}
```

and not 

```
protocol FactorGraph : BipartiteGraph {}
```

?

marcrasi:
> I think we are starting to see eye to eye more and more ;-)

Yes, this discussion is getting quite concrete and useful!

> Why does a Graph have indices, though? Can't I just have variables and factors? Do they need indices? Seems indices is an implementation detail to store the graph structure. In CS: a graph is (V, E), and a bipartite graph is (V, F, E). No indices required.

Dave can probably explain the reasons better than me, and he should be at the meeting today.

I think it has to do with the ability to stably refer to vertices in a graph even when you change data inside the vertices. (e.g. change the value of a variable). In a value-semantics world, you can't ask the language for a reference to the vertex.

This has nothing to do with the implementation of the graph storage. The concrete implementation might never store any indices.

> Also, your error function has access to all variables, but I think that's too much visibility: I just need my local variable set.

I agree that this is a semantic constraint on the protocol. The programming language is not always powerful enough to express all semantic constraints, and sometimes we might have to express the constraint as a documentation comment. Implementers are responsible for adhering to it.

I'm not 100% sure that we can't express this constraint in Swift, but I tried a bit and I currently think it's not possible to do so in a useful way. The main obstacle is that the variables can have heterogeneous types, and different factors can take different types as arguments, and (as far as I know) there's no good way to express this.

> Finally, a Swift/Protocol question.

I could go either way. I'm not sure which is best yet.

I did `associatedtype Graph` because it makes it easier on struct implementations who want to use an existing graph data structure. They can just write:
```
struct MyFactorGraph: FactorGraph {
  var graph: BipartiteAdjacencyList<X,Y,Z>
}
```

If we instead made `FactorGraph` be a refinement of `BipartiteGraph`, then the user would also have to worry about forwarding graph requirement calls to the `BipartiteAdjacencyList`. There are ways to make this really easy to do in one line, so it's not a very big deal.

marcrasi:
> So, a factor graph has additional semantics on top of (V,F,E): if a factor has a local error function , and that function has a given arity (a binary factor, say), then there better be two edges. and the type of the variables connected on the other end has to match up. Not sure how to say all this with protocols.

Ah, I didn't respond to this one.

It's the same thing about this being a semantic constraint that may not be possible to express in the programming language.

It's okay to express the constraint in the documentation.

We'll be careful to build all the other types and algorithms in the library so that it is easy to adhere to the constraint and hard to violate it.

We'll also add assertions that check this at runtime (which can be disabled when running in super optimized mode for speed).

dellaert:
> If we instead made `FactorGraph` be a refinement of `BipartiteGraph`, then the user would also have to worry about forwarding graph requirement calls to the `BipartiteAdjacencyList`. There are ways to make this really easy to do in one line, so it's not a very big deal.

A bipartite graph *is* refinement of a graph, no?

marcrasi:
> A bipartite graph is refinement of a graph, no?

Yes. Therefore, `BipartiteGraph` should probably be a protocol refinement of `Graph`.

There are some "contravariance" problems (in the sense of https://en.wikipedia.org/wiki/Covariance_and_contravariance_(computer_science)) where `MutableBipartiteGraph` is not a refinement of `MutableGraph` (https://saeta.github.io/penguin/graphs/Protocols/MutableGraph.html). Fortunately, the designers of `Graph` had enough foresight to not require mutability, so this is not a problem!

Tests/SwiftFusionTests/Inference/MetropolisTests.swift
```diff
@@ -0,0 +1,287 @@
...
+  public typealias Update = Self
+  public func updatedView(_ update: Self) -> Self {
+    return update
+  }
+}
+
+/// An example proposal to use in the test.
+extension PoseSLAMFactorGraphForSmallProposals where Pose == Pose2 {
+  /// Selects a variable uniformly at random, and some of the "nearby" variables, and applies the
+  /// same random movement to all of them.
```
ProfFan, line 172: So this is the random walk kernel Q

Sources/SwiftFusion/FactorGraph/FactorGraph.swift
```diff
@@ -0,0 +1,72 @@
...
+import PenguinGraphs
+
+/// A bipartite graph of "factor vertices" and "variable vertices", where each factor vertex is an
+/// error function over the variables that it is connected to.
+public protocol FactorGraph {
+
+  // MARK: - Graph.
+
+  /// The type of the graph.
+  associatedtype Graph: GraphProtocol
```
marcrasi, line 24: Should be a refinement of Graph.

Tests/SwiftFusionTests/Inference/MetropolisTests.swift
ProfFan: (should have a question mark here, sorry :P )

Tests/SwiftFusionTests/Inference/MetropolisTests.swift
```diff
@@ -0,0 +1,287 @@
...
+      if let update = maybeUpdate {
+        values.update(update)
+        acceptanceCount += 1
+      }
+    }
+
+    // Metropolis with this proposal is terrible for this problem, so we're not getting anywhere
+    // near the solution. So just assert that there weren't 0 or 100% acceptances so that this
+    // test tests that Metropolis did some nontrivial work at least.
+    XCTAssertGreaterThan(acceptanceCount, 0)
```
ProfFan, line 284: Maybe a good test is to start from the ground truth, and see if it works? I think metropolis with this *should* work, as this graph is pretty linear (like a Markov Chain)

Sources/SwiftFusion/FactorGraph/FactorGraph.swift
dellaert: Of BipartiteGraph :-)

Sources/SwiftFusion/FactorGraph/FactorGraph.swift
```diff
@@ -0,0 +1,72 @@
...
+  func variableIndex(of vertex: Graph.VertexId) -> Variables.Index
+
+  // MARK: - Factor error function.
+
+  /// Returns the error of the factor `factor`, at `point`.
+  ///
+  /// This can be interpreted as an unnormalized negative log probability.
+  ///
+  /// Precondition: `factor` refers to a factor vertex.
+  func error(of factor: Graph.VertexId, at point: Variables) -> Double
```
dellaert, line 57: Make clear in docs that Variables here is a *subset*. The error function also needs to know how it can get to the correct variables. e.g. in SFM we might have:
projectionError(calibration, camera_pose, 3d_point)

Sources/SwiftFusion/Inference/Metropolis.swift
```diff
@@ -0,0 +1,83 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+import PenguinGraphs
+
+/// Metropolis algorithm.
+extension FactorGraph where
+  // The Metropolis algorithm needs to query the graph for neighboring vertices:
+  Graph: IncidenceGraph,
```
dellaert, line 21: Here you actually mean incidence to variables, right? But we might also need incidence from factors to variables. I think agree with the statement from the call that a bipartite graph will almost never be used  as a “regular” graph. But one thing you can do is to calculate a regular graph on the variables *from* the factor graph.

BTW, there is a simple way to explain this in terms of sparse matrices: if a sparse matrix A encodes the non-zero pattern of a factor graph, then A^T A is the undirected (symmetric!) adjacency graph between variables, and AA^T is the adjacency graph between factors.

So, your metropolis proposal (really the only thing that wants adjacent variables) could use a *computed* adjacency graph which is a regular undirected graph, not a bipartite graph.

Tests/SwiftFusionTests/Inference/MetropolisTests.swift
marcrasi: Good idea! I'll try that when I have some time.

Sources/SwiftFusion/FactorGraph/FactorGraph.swift
marcrasi: > Make clear in docs that Variables here is a subset

Definitely.

> The error function also needs to know how it can get to the correct variables. e.g. in SFM we might have: projectionError(calibration, camera_pose, 3d_point)

The _implementation_ of the error function needs to know how to get the correct variables. The _interface_ of the error function does not need to specify anything about how that happens. This is a very interesting point that we can talk about at our next call!

Sources/SwiftFusion/Inference/Metropolis.swift
marcrasi: The metropolis proposal might want variable adjacency information, yes. (Or it might not, depending on what the proposal is.)

However, this implementation of metropolis is completely agnostic to the proposal. This implementation just takes the proposal as a closure.

This implementation of metropolis uses the graph structure to determine which factors must be recomputed to calculate the acceptance ratio. Specifically, it looks at the proposal, sees which variables have changed, and only recomputes factors adjacent to those variables.

This implementation is only one point in the design space of generic metropolis algorithms. It could be a totally wrong point in the design space. Perhaps the proposal should be responsible for calculating the set of affected factors, so that the generic metropolis algorithm doesn't need to inspect the graph structure itself.

Tests/SwiftFusionTests/Inference/MetropolisTests.swift
marcrasi: Could be? I don't know what that is.

I just came up with an arbitrary proposal that does something interested with the graph structure, to illustrate possibilities. It might be a pretty bad one. Or it might have a name?

marcrasi:
Closing this to clean up, since it won't be relevant with the new generic factor graph data structure.

PR #61, opened 2020-05-18
benchmarking infrastructure

marcrasi:
This adds a benchmark for Pose2SLAM using NonlinearFactorGraph, Gauss-Newton, and CGLS.

I added a section of the readme showing how to run it.

Since this adds an external dependency, I committed a `Package.resolved` that pins us to a specific commit. `swift package update` updates this to newer commits.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,6 @@
 /Packages
 /*.xcodeproj
 xcuserdata/
-/Package.resolved
 /.swiftpm
 /.vscode
 /.idea
--- a/Sources/Benchmarks/Pose2SLAM.swift
+++ b/Sources/Benchmarks/Pose2SLAM.swift
@@ -1,0 +1,70 @@
+// Copyright 2020 The SwiftFusion Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/// Benchmarks Pose2SLAM solutions.
+
+import Benchmark
+import SwiftFusion
+
+let pose2SLAM = BenchmarkSuite(name: "Pose2SLAM") { suite in
+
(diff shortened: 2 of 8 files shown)


Sources/Benchmarks/Pose2SLAM.swift
```diff
@@ -0,0 +1,70 @@
...
+  let intelDataset = try! G2OFactorGraph(fromG2O: try! cachedDataset("input_INTEL_g2o.txt"))
+  check(intelDataset.graph.error(intelDataset.initialGuess), near: 73565.64, accuracy: 1e-2)
+
+  // Uses `NonlinearFactorGraph` on the Intel dataset.
+  // The solvers are configured to run for a constant number of steps.
+  // The nonlinear solver is 5 iterations of Gauss-Newton.
+  // The linear solver is 100 iterations of CGLS.
+  suite.benchmark(
+    "NonlinearFactorGraph, Intel, 5 Gauss-Newton steps, 100 CGLS steps",
+    settings: .iterations(1)
```
shabalind, line 31: Results are going to be quite noisy if you only run a single iteration. Is it not practical to run it for longer?

Sources/Benchmarks/Pose2SLAM.swift
```diff
@@ -0,0 +1,70 @@
...
+  ) {
+    var val = intelDataset.initialGuess
+    for _ in 0..<5 {
```
shabalind, line 34: Usually, it's better to avoid loops in benchmarks, only put the loop body, and then customize number of iterations with `.iterations` (or better leave it off to be automatically detected by the framework).

Sources/Benchmarks/Pose2SLAM.swift
marcrasi: This is benchmarking a whole solver, and it's pretty slow right now (~12 seconds).

Sources/Benchmarks/Pose2SLAM.swift
marcrasi: This loop is part of the solver logic that we are benchmarking, so I think it makes sense to have it in the benchmark. Eventually we're going to have an API for the solver and the benchmark body will look like `solve(graph, initialGuess)` and this loop will be hidden inside the `solve` function.

ProfFan: Also in the final solver this should be a `while error < tol || i < max_iteration`

Sources/Benchmarks/Pose2SLAM.swift
shabalind: Thanks for the clarification.

.gitignore
```diff
@@ -3,7 +3,6 @@
 /Packages
 /*.xcodeproj
 xcuserdata/
-/Package.resolved
```
ProfFan, line 6: Should this be checked in?

PR #64, opened 2020-05-20
vector cleanup and differentiable vector operations

marcrasi:
The main improvements in this PR are:
* All vector operations are now differentiable.
* You can do scalar multiplication with `s * v`. The old way `v.scaled(by: s)` doesn't exist any more.

Full details:
* Renames `EuclideanVectorSpace` to `EuclideanVector` for consistency with other Swift protocol names. They are supposed to name instances of conforming types, so "Vector" is better than "VectorSpace". (This is like `Differentiable.TangentVector`.)
* Removes all use of `VectorProtocol` because it's an unprincipled protocol that causes trouble and that we'll probably remove from the stdlib. (This removes the `v.scaled(by: s)` thing and frees us up to do it with `s * v` instead!)
* Makes all the `EuclideanVector` operations differentiable, for all types that conform to `EuclideanVector`.
* Adds a general `EuclideanVector` test that tests all types that conform to `EuclideanVector`.
* Generates `VectorN` up to `Vector10`, and deletes the definition of `Vector6` in `Pose3.swift`.


dellaert:
@ProfFan @marcrasi I propose the following policy: authors merge their own PR after approval, and also delete the branch if they do not plan to work on it anymore. 

Rationale: might want to add some uncontroversial finishing touches after review (spelling, TODO, comments), and we also don't want a lot of stale branches - a delete button is right there after merging. But I'd hesitate deleting someone else's branch.

ProfFan:
@dellaert I fully agree, will do in the future :)

dellaert:
> @dellaert I fully agree, will do in the future :)

Cool. Both please delete branches you created here: https://github.com/borglab/SwiftFusion/branches/active

PR #65, opened 2020-05-22
simplify vector protocols

marcrasi:
We have a large number of vector-related protocols: `EuclideanVector`, `FixedDimensionVector`, `TangentStandardBasis`, `VectorConvertible`, `TensorConvertible`. This makes things pretty confusing to work with.

This PR simplifies things into 2 protocols:
* `EuclideanVector`: A Euclidean vector space.
* `EuclideanVectorN`: A refinement of `EuclideanVector` that requires compiletime known dimension.

One bad consequence of this change is that `BearingRangeError` now requires a lot more boilerplate. I have some separate ideas that could eliminate all of that boilerplate, so this should not be a problem for long.


dellaert:
So, when I see the word "EuclideanVector" my mind immediately goes to "Why Euclidean", esp. in light of our discussion that n*n skew-symmetric matrices form a vector space, as well.

Quoting from https://en.wikipedia.org/wiki/Tangent_space

> In differential geometry, one can attach to every point x of a differentiable manifold a tangent space—a real vector space that intuitively contains the possible directions in which one can tangentially pass through x. The elements of the tangent space at x are called the tangent vectors at x. This is a generalization of the notion of a bound vector in a Euclidean space. The dimension of the tangent space at every point of a connected manifold is the same as that of the manifold itself.

But, we of course only need to resolve this issue when it will practically matter , i.e. some unit tests I want to see run, fails because of this.

PR #66, opened 2020-05-22
new group operators

marcrasi:
I propose that we use something other than `*` for the group operation and the group action. I have no strong opinion on what other operator we should use, please suggest better options! (Is there any precedent in other libraries?)

Reasons:

`*` will be highly confusing to human readers when we make `VectorN` conform to `LieGroup`, because `a * b` looks like some kind of multiplication but it's actually `a + b`!!

The Swift typechecker is notoriously slow when it has to resolve overloaded operators. This simple change hugely decreases compile time.

Before:
```
marcrasi@marcrasi:~/git/SwiftFusion$ swift package clean && time swift build
[88/88] Linking Benchmarks

real    0m27.196s
user    1m54.372s
sys     0m5.155s
```

After:
```
marcrasi@marcrasi:~/git/SwiftFusion$ swift package clean && time swift build
[88/88] Linking Benchmarks

real    0m18.016s
user    1m47.860s
sys     0m4.806s
```

### DIFF
--- a/Sources/SwiftFusion/Core/LieGroup.swift
+++ b/Sources/SwiftFusion/Core/LieGroup.swift
@@ -1,5 +1,11 @@
 import TensorFlow
 
+/// The group operation of a Lie group.
+infix operator **: MultiplicationPrecedence
+
+/// The group action of a Lie group on a set.
+infix operator *^: MultiplicationPrecedence
+
 /// An element of a Lie group.
 ///
 /// To create a Lie group type, follow the manifold recipe [1] and conform the coordinate type to
@@ -39,18 +45,18 @@ extension LieGroup {
   @differentiable(wrt: (lhs, rhs))
   @differentiable(wrt: lhs)
   @differentiable(wrt: rhs)
-  public static func * (_ lhs: Self, _ rhs: Self) -> Self {
-    return Self(coordinate: lhs.coordinate * rhs.coordinate)
+  public static func ** (_ lhs: Self, _ rhs: Self) -> Self {
+    return Self(coordinate: lhs.coordinate ** rhs.coordinate)
   }
 
-  /// Derivative of `*` with respect to both sides.
+  /// Derivative of `**` with respect to both sides.
   ///
   /// We define separate derivatives with respect to different sets of arguments so that we can
   /// avoid doing unnecessary work when some of the arguments are not changing.
   ///
   /// Swift AD can compute this, but we know a mathematical expression for the derivative that
   /// is more efficient than the compiler can figure out.
-  @derivative(of: *, wrt: (lhs, rhs))
+  @derivative(of: **, wrt: (lhs, rhs))
   @usableFromInline
   static func vjpGroupOperationWrtBoth(_ lhs: Self, _ rhs: Self) ->
     (value: Self, pullback: (TangentVector) -> (TangentVector, TangentVector))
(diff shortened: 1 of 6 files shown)


dellaert:
Well, multiplying Rot3 is definitely very intuitive with `*`. I always complain loudly when numpy forces me to use `@` for matrix multiplication. Is it possible to just add a protocol conformance that implements * as compose for when we want it? And hence, choose the operator based on the concrete type?

marcrasi:
> Is it possible to just add a protocol conformance that implements * as compose for when we want it? And hence, choose the operator based on the concrete type?

Yes, something like that would be possible but I think it would be pretty confusing to have different names for the same generic operation depending on what the concrete type is. So from a human readability perspective, I'd prefer to always use `*` over doing something like that.

As for the compile speed problem, there are ways to hint to the compiler what the correct overload resolution is, so we can increase the compile speed without changing the operators.

dellaert:
I actually think from a human perspective we would want to choose the operator on a per case basis. Vectors add, although they are *technically* a Lie group.  But adding rotation matrices: makes no sense !

Note that skew symmetric matrices are an interesting case: as a vector space we'd use +, but to multiply with a vector (which for 3x3 implements cross product) we would definitely use *.

marcrasi:
Maybe the right thing is to have a special operator for generic lie group compose, and simply define it in terms of concrete operations of each of the conforming types? Like this:

```swift
protocol LieGroup {
  // Generic compose operator that works on all lie groups.
  static func ** (_ lhs: Self, _ rhs: Self) -> Self
}

struct Vector3: LieGroup {
  /// Vector addition.
  static func + (_ lhs: Self, _ rhs: Self) -> Self { ... }

  /// Lie group composition is vector addition.
  static func ** (_ lhs: Self, _ rhs: Self) -> Self { return lhs + rhs }
}

struct Rot3: LieGroup {
  /// Rotation composition.
  static func * (_ lhs: Self, _ rhs: Self) -> Self { ... }

  /// Lie group composition is rotation composition.
  static func ** (_ lhs: Self, _ rhs: Self) -> Self { return lhs * rhs }

  /// Action on a vector.
  static func * (_ lhs: Self, _ rhs: Vector3) -> Vector3 { ... }
}
```

dellaert:
I like everything here *except* `**` :-) Why do we need it?

marcrasi:
Are you suggesting that we name it something else (`*`?) or that we completely remove the concept of a generic lie group composition that works on any lie group type?

We currently use a generic lie group composition operator in some places, like `BetweeFactor`, so that `BetweenFactor` can work generically on any lie group type: https://github.com/borglab/SwiftFusion/blob/37069b458565988d8a51a118c7e4ff9e3648f2c9/Sources/SwiftFusion/Inference/BetweenFactor.swift#L70

dellaert:
Well, I see operators as convenience for users, so if not used in too many places I'd indeed remove `**` as an operator and just use `compose` for generic Lie group code, that is agnostic as to whether group is additive or multiplicative.

PS See also [Group.h](https://github.com/borglab/gtsam/blob/f538d1dc7bdd7126cb683f3e961c985f76a872b0/gtsam/base/Group.h) in GTSAM. You'll see we have AdditiveGroup and MultiplicativeGroup traits.

ProfFan:
My two cents on this is that I personally liked the `*` notation, as this eliminates the need for nested `.compose(X.compose(...))` calls, but I see Frank's point. Maybe `**` could be a solution?

marcrasi:
> Well, I see operators as convenience for users, so if not used in too many places I'd indeed remove ** as an operator and just use compose for generic Lie group code, that is agnostic as to whether group is additive or multiplicative.

I see, that makes a lot of sense! I think we should do this.

> PS See also Group.h in GTSAM. You'll see we have AdditiveGroup and MultiplicativeGroup traits.

Is there anything fundamentally difference between an abstract "AdditiveGroup" and an abstract "MultiplicativeGroup"? Something like an operation that makes sense to define on an abstract additive group but not on an abstract multiplicative group? If so, then I see that having something like `protocol AdditiveGroup` and `protocol MultiplicativeGroup` would be useful, and these protocols could automatically define `compose` in terms of `+` or `*`.

However, if there's no difference, then I think it's better to explicitly define `compose` for each concrete type because in this case `protocol AdditiveGroup` and `protocol MultiplicativeGroup` don't serve any purpose other than adding a bit of magic that saves a bit of typing.

In any case, all of this interacts with the AD stuff nontrivially so it's not completely obvious how to set it up to all work, so I won't do anything immediately. I'll just keep it in my head and see when it makes sense to make some progress on it.

ProfFan:
> Is there anything fundamentally different between an abstract "AdditiveGroup" and an abstract "MultiplicativeGroup"? Something like an operation that makes sense to define on an abstract additive group but not on an abstract multiplicative group? If so, then I see that having something like `protocol AdditiveGroup` and `protocol MultiplicativeGroup` would be useful, and these protocols could automatically define `compose` in terms of `+` or `*`.

A convention is that additive groups are usually Abelian (i.e. `a+b=b+a`), but I don't know if that is a wide-enough consensus. Other than that, I see no difference as the operator symbol `+` or `*` is just an alias of the "group operation".

ProfFan:
> Something like an operation that makes sense to define on an abstract additive group but not on an abstract multiplicative group?

For abelian groups the adjoint (`g^{-1} * x * g`) (inner automorphism in group theory) is the identity (trivial).

dellaert:
> > Something like an operation that makes sense to define on an abstract additive group but not on an abstract multiplicative group?
> 
> For abelian groups the adjoint (`g^{-1} * x * g`) (inner automorphism in group theory) is the identity (trivial).

Good find!!!!

ProfFan:
I still tend to vote for `**`, but @dellaert should give our final verdict?

dellaert:
> I still tend to vote for `**`, but @dellaert should give our final verdict?

I associated `**` with powers. *if* you have to have an operator, perhaps `@`.

marcrasi:
Closing for now, since there's nothing concrete being blocked by this.

Issue #52, opened 2020-05-12
Replace Tensor<Double> with Matrix3x3 in Rot3


marcrasi:
Matrix3x3 is faster because it will have no heap allocation and no TensorFlow overhead.

As mentioned in https://github.com/borglab/SwiftFusion/issues/42#issuecomment-627413240, this is currently the main bottleneck for Pose3Slam.

ProfFan:
As #68 is merged, the next step would be making the generic version of MatrixN

ProfFan:
@marcrasi Could you change the G2O Reader a bit and add the ability to read in the Pose3 dataset [here](https://github.com/HeYijia/GraphSLAM_tutorials_code/blob/master/g2o_test/data/sphere_bignoise_vertex3.g2o)? I tried and it seems to be a bit too convoluted for me...

dellaert:
tl;dr As commented on in PR #68 I believe there is value in seeing a matrix as an array of column vectors. You can also see it as an array of co-vectors corresponding to rows, but that is less common. matvec is then taking the linear combination of the columns.

Long version:

Because I like types, I actually prefer to think about tensors, where each dimension can be "contravariant" or "covariant". A column vector is a 1D contravariant vector, and a row vector is a 1D covariant vector.  Things like 3D-position etc are naturally "contra-variant": if you transform the coordinate frame points transform in the opposite way. A matrix as we know it has both a contravariant index i and a covariant index j, e.g., j=0 extracts the first column.

You can only "contract" - a generalization of dot product - covariant with contravariant dimensions. This is the "[Einstein notation](https://en.wikipedia.org/wiki/Einstein_notation)": dot = p_i q^i. 

What is matrix multiplication then? Using Einstein notation, we have A^i_k B^k_j = C^i_j, and so we can see that A^i_k is a bunch of column vectors a^i, indexed by k, and B^k_j is a bunch of weight column vectors b^k, indexed by j. 

In physics you can mix and match co- and contravariant indices, and the order does not matter as they have symbols associated. But in an implementation you do care about the order, and so if we just implement Matrix here then you have to choose: an array of column vectors, or an array of co-vectors. Because we typically think of vectors as column vectors, the former is more natural and agrees with common sense.

Note there is one place where co-vectors show up all the time: automatic differentiation :-) The tangent vectors are column vectors, but adjoints I believe are typically seen as co-vectors. I forget where I saw that.

[...]

PR #68, opened 2020-05-24
Matrix3 for Rot3

ProfFan:
This implements the static 3x3 matrix version of Rot3. Pose3SLAM now runs 10x faster :)

### DIFF
--- a/Sources/SwiftFusion/Core/Matrix.swift
+++ b/Sources/SwiftFusion/Core/Matrix.swift
@@ -6,11 +6,12 @@ public struct Matrix: Equatable {
   public fileprivate(set) var scalars: [Double]
 
   /// The number of rows and columns in the matrix.
+  @noDerivative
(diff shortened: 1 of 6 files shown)


dellaert:
Awesome, *but* I think things will be a lot simpler if you implement Matrix3 as 3 Vector3 columns?

ProfFan:
@dellaert I think that is also a viable solution, but I believe sooner or later this would be switched to a new fixed-size matrix implementation by Marc and Dave, so let's keep this for now.

dellaert:
> @dellaert I think that is also a viable solution, but I believe sooner or later this would be switched to a new fixed-size matrix implementation by Marc and Dave, so let's keep this for now.

Sure! For that implementation, then, try to keep in mind that a Matrix is best seen as an array of column vectors. You can also see it as an array of co-vectors corresponding to rows, but that is less common. I'll comment on issue #52 with some thoughts.

PR #69, opened 2020-05-25
g2o 3d parser

marcrasi:
Mostly updates the G2O reader to read 3d files.

I don't know how to convert quaternions to rotation matrices, so I added a dummy implementation of that. @ProfFan would you be able to fill in the real implementation, at G2OReader.swift:184, and in the test?


ProfFan:
Issues are emerging when testing with the big dataset (sphere), which is good news :)

1. the current solver does not converge (which is somewhat expected as the error goes up in CGLS if we iterate too much cycles)
2. we cannot test against GTSAM as there is no noise model support

ProfFan:
Added a termination condition for CGLS when the residual is growing instead of decreasing, and it seems to be working for the big dataset. Waiting for the final result.

ProfFan:
```
error = 223181364.32139167
gfg error = 20876073.157318663
error = 410459970.7309142
gfg error = 225426441.0518634
error = -nan
```
CGLS is diverging, need to investigate why

ProfFan:
@marcrasi Please review :) We can deal with the issues later

marcrasi:
@ProfFan, your changes look good to me. (I can't "approve" this PR because I'm the author:))

Sources/Benchmarks/Pose2SLAM.swift
```diff
@@ -41,24 +42,7 @@ let pose2SLAM = BenchmarkSuite(name: "Pose2SLAM") { suite in
...
-  /// The factor graph representing the measurements.
-  var graph: NonlinearFactorGraph = NonlinearFactorGraph()
-
-  public mutating func addInitialGuess(index: Int, pose: Pose2) {
-    initialGuess.insert(index, pose)
-  }
-
-  public mutating func addMeasurement(frameIndex: Int, measuredIndex: Int, pose: Pose2) {
-    graph += BetweenFactor(frameIndex, measuredIndex, pose)
+//    check(intelDataset.graph.error(val), near: 35.59, accuracy: 1e-2)
```
ProfFan, line 45: I commented out this check because now it's around 20

Sources/SwiftFusion/Datasets/G2OReader.swift
```diff
@@ -14,118 +14,200 @@
...
+      let firstColumn = try parser.parseString()
+      if firstColumn == "VERTEX_SE3:QUAT" {
+        handleEntry(.initialGuess(index: try parser.parseInt(), pose: try parser.parsePose3()))
+      } else if firstColumn == "EDGE_SE3:QUAT" {
+        handleEntry(.measurement(
+          frameIndex: try parser.parseInt(),
+          measuredIndex: try parser.parseInt(),
+          pose: try parser.parsePose3()
+        ))
+        // Consume the information matrix columns.
```
ProfFan, line 102: note: 21 entrys of the upper triangle of the information matrix

Tests/SwiftFusionTests/Datasets/G2OReader.swift
```diff
@@ -33,19 +35,50 @@ final class G2OReaderTests: XCTestCase {
     XCTAssertEqual(dataset, expectedDataset)
   }
 
+  /// Tests loading a simple 3d g2o file using G2OReader.
+  func testLoadSimple3D() {
+    /// Returns the expected `Pose3` that should be read from `g2oEntries`.
+    func pose3(_ g2oEntries: [Double]) -> Pose3 {
+      let s = g2oEntries
+      let t = Vector3(s[0], s[1], s[2])
+      let r = Rot3.fromQuaternion(s[6], s[3], s[4], s[5])
```
ProfFan, line 44: w, x, y, z but g2o is x, y, z, w, thus above

future note: should make this function use labels `fromQuat(w:, x:, y:, z:)` to avoid typos

PR #70, opened 2020-05-29
fixed size vectors and matrices

marcrasi:
Here is some initial work on making principled fixed size vectors and matrices.

Everything is stored as a `FixedSizeArray`:
* Vectors are `FixedSizeArray`s of vectors (with a base case `ScalarVector` type).
* Matrices are `FixedSizeArray`s of linear maps (with a base case that a single vector is the linear map of scalar multiplication by that vector).

As a result, matrices are arrays of columns and matrix multiplication works by scaling each column by the corresponding scalar, like @dellaert suggested in https://github.com/borglab/SwiftFusion/issues/52#issuecomment-633286415.

I put this in a new target so that I could use names like `Vector` and `Matrix` without clashing with the existing types. My goal is to build all the required features in here and then eventually start using these new types in the rest of SwiftFusion.


Tests/LinearAlgebraTests/LinearAlgebraTests.swift
```diff
@@ -0,0 +1,46 @@
+import LinearAlgebra
+import XCTest
+
+final class LinearAlgebraTests: XCTestCase {
+  /// Tests adding two vectors.
+  func testVectorAdd() {
+    let v1 = Vector2d(1, 2)
```
marcrasi, line 7: "2d" is confusing in the typealias.

Vector2d = Vector2<Double>

Tests/LinearAlgebraTests/LinearAlgebraTests.swift
```diff
@@ -0,0 +1,46 @@
...
+      Vector3d(1, 0, 1),
+      Vector3d(0, 1, 1)
+    )
+    let v = Vector2d(2, 3)
+    XCTAssertEqual(mat(v), Vector3d(2, 3, 5))
+  }
+
+  /// Tests multiplying two matrices.
+  func testMatmul() {
+    let mat1 = Matrix3x2d(
```
marcrasi, line 30: consider making it possible to do

Matrix(Vector(1, 0, 1), Vector(0, 1, 1))

Tests/LinearAlgebraTests/LinearAlgebraTests.swift
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  /// Tests applying a matrix to a vector.
+  func testMatrixVector() {
+    let mat = Matrix3x2d(
+      Vector3d(1, 0, 1),
+      Vector3d(0, 1, 1)
+    )
+    let v = Vector2d(2, 3)
+    XCTAssertEqual(mat(v), Vector3d(2, 3, 5))
```
marcrasi, line 25: consider `mat * v`

Tests/LinearAlgebraTests/LinearAlgebraTests.swift
```diff
@@ -0,0 +1,46 @@
...
+      Vector2d(2, 3),
+      Vector2d(10, 10)
+    )
+
+    let expectedProduct = Matrix3x2d(
+      Vector3d(2, 3, 5),
+      Vector3d(10, 10, 20)
+    )
+
+    XCTAssertEqual(mat1.matmul(mat2), expectedProduct)
```
marcrasi, line 44: mat1 * mat2

Sources/LinearAlgebra/FixedSizeArray+Vector.swift
```diff
@@ -0,0 +1,35 @@
+/// A fixed-size array of `Vector`s is a `Vector`.
+extension FixedSizeArray where Element: Vector {
+  public typealias Scalar = Element.Scalar
+  public mutating func add(_ other: Self) {
+    for i in indices {
```
marcrasi, line 5: look if loops are getting unrolled

Sources/LinearAlgebra/FixedSizeArray+Vector.swift
```diff
@@ -0,0 +1,35 @@
...
+extension Array2: AdditiveArithmetic, Vector where Element: Vector {}
+extension Array3: AdditiveArithmetic, Vector where Element: Vector {}
+
+extension Array1: LinearMap where Element: Vector {}
+extension Array2: LinearMap where Element: Vector {}
+extension Array3: LinearMap where Element: Vector {}
+
+public typealias Vector1<Scalar> = Array1<ScalarVector<Scalar>>
+public typealias Vector2<Scalar> = Array2<ScalarVector<Scalar>>
+public typealias Vector3<Scalar> = Array3<ScalarVector<Scalar>>
```
marcrasi, line 31: alternative:
```
Vector<Array2<Double>>
Vector<Array2<Array2<Double>>>
```

marcrasi:
Closing because this turned out not to be important for the initial generic factor graph work. We should revisit eventually.

PR #72, opened 2020-05-29
covectors and differentiable vectors

marcrasi:
This adds differentiability to the vector abstractions in https://github.com/borglab/SwiftFusion/pull/70.

We need covectors to define pullbacks of vector space operations, so I also added covectors.


dellaert:
All checks failed? :-)

ProfFan:
@dellaert Looks like a GitHub issue, I restarted the build

ProfFan:
@dellaert Test is successful

marcrasi:
Closing because this turned out not to be important for the initial generic factor graph work. We should revisit eventually.

PR #75, opened 2020-06-05
generic linearizable and gaussian factors

marcrasi:
This adds refined factor protocols "GenericLinearizableFactor" and "GenericGaussianFactor".

The main point is to allow solvers that do iterated linearization. The solver linearizes all the "GenericLinearizableFactors" at the current point to get a bunch of "GenericGaussianFactors", and then it solves the factor graph of "GenericGaussianFactors". (No solver included in this PR).

### DIFF
--- a/Sources/SwiftFusion/Core/Tuple+EuclideanVectorN.swift
+++ b/Sources/SwiftFusion/Core/Tuple+EuclideanVectorN.swift
@@ -1,0 +1,157 @@
+// Copyright 2020 The SwiftFusion Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import PenguinStructures
+
+extension Empty: AdditiveArithmetic {
+  public static func + (_ lhs: Self, _ rhs: Self) -> Self {
+    return Empty()
+  }
+  public static func - (_ lhs: Self, _ rhs: Self) -> Self {
+    return Empty()
+  }
+  public static var zero: Self {
+    return Empty()
+  }
+}
+
+extension Tuple: AdditiveArithmetic where Head: AdditiveArithmetic, Tail: AdditiveArithmetic {
+  public static func + (_ lhs: Self, _ rhs: Self) -> Self {
+    return Tuple(head: lhs.head + rhs.head, tail: lhs.tail + rhs.tail)
+  }
+  public static func - (_ lhs: Self, _ rhs: Self) -> Self {
+    return Tuple(head: lhs.head - rhs.head, tail: lhs.tail - rhs.tail)
+  }
+  public static var zero: Self {
+    return Tuple(head: Head.zero, tail: Tail.zero)
+  }
+}
+
+extension Empty: Differentiable {
+  public typealias TangentVector = Self
+  public mutating func move(along direction: TangentVector) {}
+}
+
+extension Tuple: Differentiable
+where Head: Differentiable, Tail: Differentiable, Tail.TangentVector: TupleProtocol {
+  public typealias TangentVector = Tuple<Head.TangentVector, Tail.TangentVector>
+  public mutating func move(along direction: TangentVector) {
+    head.move(along: direction.head)
+    tail.move(along: direction.tail)
+  }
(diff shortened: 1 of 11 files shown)


marcrasi:
This has become a bit stale because of https://github.com/borglab/SwiftFusion/pull/77 and https://github.com/borglab/SwiftFusion/pull/78, so I have marked this as draft until I freshen it.

marcrasi:
This is now freshened and ready for review from @dabrahams .

Sources/SwiftFusion/Core/Tuple+EuclideanVectorN.swift
```diff
@@ -0,0 +1,159 @@
...
+  /// Creates an instance with the same elements as `vector`.
+  @differentiable
+  public init(_ vector: Vector) { self.init() }
+
+  /// A `Vector` containing the same elements as `self`.
+  @differentiable
+  public var vector: Vector { return Vector([]) }
+}
+
+extension Empty: EuclideanVectorN {
```
marcrasi, line 84: This `EuclideanVectorN` is the "fixed size vector" protocol that exists in SwiftFusion now. It has some problems, but I think it's good enough to get the generic factor graph stuff working, so I'm going to use it in all the generic factor graph stuff and we can replace it later when we have more time to figure out fixed size vectors.

Sources/SwiftFusion/Core/Tuple+EuclideanVectorN.swift
```diff
@@ -0,0 +1,159 @@
...
+  }
+  public static func - (_ lhs: Self, _ rhs: Self) -> Self {
+    return Empty()
+  }
+  public static var zero: Self {
+    return Empty()
+  }
+}
+
+extension Tuple: AdditiveArithmetic where Head: AdditiveArithmetic, Tail: AdditiveArithmetic {
```
dabrahams, line 29: Not a definitive thing whatsoever, just a note:  I'm a little wary about conditionally conforming `Tuple` to every possible thing for the same reasons I don't think `Array<Layer>` should conform to `Layer`.  It's kind of sad that we don't have scoped conformances, because that would make this much less of an issue.

Sources/SwiftFusion/Inference/FactorsStorage.swift
```diff
@@ -71,3 +73,248 @@ final class FactorArrayStorage<Element: GenericFactor>:
...
+  func linearAdjoint_(_ errorVectorsStart: UnsafeRawPointer, into result: inout ValuesArray)
+}
+
+/// APIs that depend on `GaussianFactor` `Element` type.
+extension ArrayStorageImplementation where Element: GenericGaussianFactor {
+  /// Returns the results of the factors' linear functions at the given point.
+  func linearForward(_ variableAssignments: ValuesArray)
+    -> ArrayBuffer<VectorArrayStorage<Element.ErrorVector>>
+  {
+    Element.Variables.withVariableBufferBaseUnsafePointers(variableAssignments) { varsBufs in
```
dabrahams, line 243: Seems like `variableAssignments` should be the receiver here.

Sources/SwiftFusion/Core/Tuple+EuclideanVectorN.swift
```diff
@@ -0,0 +1,159 @@
...
+    return head.dot(other.head) + tail.dot(other.tail)
+  }
+
+  // MARK: - Conversion to/from a concrete dynamically-sized `Vector`.
+
+  /// Creates an instance with the same elements as `vector`.
+  @differentiable
+  public init(_ vector: Vector) {
+    self.init(
+      head: Head(Vector(Array(vector.scalars[0..<Head.dimension]))),
```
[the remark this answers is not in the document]
marcrasi: I have improved this. Hopefully it should be clearer now. It did require me to make some changes to `EuclideanVectorN`, but it's worth it because I'm going to need these changes in an upcoming PR.

Sources/SwiftFusion/Inference/GenericFactor.swift
```diff
@@ -29,13 +29,42 @@ protocol GenericFactor {
...
+
+  /// Returns a factor whose `errorVector` linearly approximates `self`'s `errorVector` around the
+  /// given point.
+  func linearized(_: Variables) -> Linearized
+}
+
+/// A factor with a vector error that is a linear function of the input, plus a constant.
+protocol GenericGaussianFactor: GenericLinearizableFactor where Variables: EuclideanVector {
+  /// Returns the result of the linear function at the given point.
+  func linearForward(_: Variables) -> ErrorVector
```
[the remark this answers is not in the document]
dellaert: All, I am *clobbered* this week due to a paper deadline (as I communicated two weeks ago) so I really don't have the time to get the context at this point :-/ Next week I can spend some quality time with swift.

marcrasi:
I added TODOs for most things, and I plan to do the suggested renamings in a later PR after the changes have settled down.

marcrasi:
Oops, meant to merge not close.

PR #77, opened 2020-06-05
change generic factor graph example to SwitchingMotionModelFactor

marcrasi:


Tests/SwiftFusionTests/Inference/GenericFactorTests.swift
```diff
@@ -19,57 +19,87 @@ import XCTest
...
+  typealias Variables = Tuple3<Int, Pose, Pose>
+
+  let edges: Variables.Indices
+
+  let motions: [Pose]
+
+  func error(_ variables: Variables) -> Double {
+    let (motionLabel, start, end) = (variables.head, variables.tail.head, variables.tail.tail.head)
+    let actualMotion = between(start, end)
+    return motions[motionLabel].localCoordinate(actualMotion).squaredNorm
```
dellaert, line 47: Interesting that we do squaredNorm. This is a departure from GTSAM, where we simply return the error vector. We should probably discuss this in more detail.

Tests/SwiftFusionTests/Inference/GenericFactorTests.swift
```diff
@@ -19,57 +19,87 @@ import XCTest
...
   func error(_ variables: Variables) -> Double {
-    let (priorIndex, position) = (variables.head, variables.tail.head)
-    return priors[priorIndex].localCoordinate(position).squaredNorm
+    let actualPosition = variables.head
+    return prior.localCoordinate(actualPosition).squaredNorm
+  }
+}
+
+/// A factor that switches between different linear motions based on an integer label.
+fileprivate struct SwitchingLinearMotionFactor<Pose: LieGroup>: GenericFactor {
```
dellaert, line 37: Love it! But I think this is not a linear model. So, `SwitchingMotionModelFactor` ?

Tests/SwiftFusionTests/Inference/GenericFactorTests.swift
```diff
@@ -19,57 +19,87 @@ import XCTest
...
+    // The factors are:
+    // - a prior factor giving a prior of `Pose2(0, 0, 0)` on the first position.
+    // - two motion factors with possible motions [Pose2(1, 1, 0), Pose2(0, 0, 1)]
+
+    // Set up the initial guess.
 
     let intVariables = ArrayStorage<Int>.create(minimumCapacity: 2)
-    let intVar1 = TypedID<Int, Int>(intVariables.append(0)!)
-    let intVar2 = TypedID<Int, Int>(intVariables.append(1)!)
+    let motionLabel1ID = TypedID<Int, Int>(intVariables.append(0)!)
```
dellaert, line 67: TypedID is so cool! But, wow, this syntax is a bit heavy

Tests/SwiftFusionTests/Inference/GenericFactorTests.swift
```diff
@@ -19,57 +19,87 @@ import XCTest
...
+    let priorFactors = FactorArrayStorage<PriorFactor<Pose2>>.create(minimumCapacity: 1)
+    _ = priorFactors.append(PriorFactor(edges: Tuple1(pose1ID), prior: Pose2(0, 0, 0)))
+
+    let motionFactors =
+      FactorArrayStorage<SwitchingLinearMotionFactor<Pose2>>.create(minimumCapacity: 2)
+    _ = motionFactors.append(SwitchingLinearMotionFactor(
+      edges: Tuple3(motionLabel1ID, pose1ID, pose2ID),
+      motions: [Pose2(1, 1, 0), Pose2(0, 0, 1)]
+    ))
+    _ = motionFactors.append(SwitchingLinearMotionFactor(
```
dellaert, line 91: no such thing as "emplace" I guess?

Tests/SwiftFusionTests/Inference/GenericFactorTests.swift
marcrasi: Done

Tests/SwiftFusionTests/Inference/GenericFactorTests.swift
marcrasi: This returns `error` instead of `errorVector` because the protocol for factors with error vectors doesn't exist yet in feature/generic_factor_graph. That protocol is implemented in https://github.com/borglab/SwiftFusion/pull/75. After I merge this PR, I'll update https://github.com/borglab/SwiftFusion/pull/75 and you'll be able to see the `errorVector` methods for these factors.

Tests/SwiftFusionTests/Inference/GenericFactorTests.swift
marcrasi: This will be improved by one of the pieces that we haven't implemented yet -- there will be some container for variable assignments and adding a new variable will look something like this:

```
let motionLabel1ID = variableAssignments.store(0)
```

Tests/SwiftFusionTests/Inference/GenericFactorTests.swift
marcrasi: No such thing, and I think Swift's value semantics make it so that "append" is just as fast as "emplace" would be. Though I'm not sure, because I'm not very knowledgeable about value semantics and "emplace". @dabrahams might know for sure.

dellaert: The feature of emplace that I like is that we don't have to repeat the constructor here, rather the constructor arguments are forwarded, yielding less clutter - the collection knows the type, yes?

marcrasi: Yeah, that doesn't exist, at least not automatically.

There is a way to manually define `emplace` for collections that know their element type, but you have to write separate code for each element type:

```
extension RangeReplacableCollection {
  func emplace<Pose: LieGroup>(edges: Element.Edges, motions: [Pose]) where Element == SwitchingMotionModelFactor<Pose> {
    append(SwitchingMotionModelFactor(edges: edges, motions: motions)
  }
}
```

Probably not worth it if we have to do it for each factor? Anyways, I'm not going to do anything about this now because I'm focusing on getting the functionality working without any sugar.

dabrahams: You can just replace `SwitchingLinearMotionFactor` with `.init`.

PR #79, opened 2020-06-11
Pose2SLAM benchmark with GenericFactorGraph

marcrasi:
Adds:
* `VariableAssignments` methods that make it nicer and easier to use.
* `GenericFactorGraph` and `GenericGaussianFactorGraph` with high-level methods that make it easy to build a graph and linearize it.
* `GenericJacobianFactor` for linearizing factors.
* `GenericPriorFactor` and `GenericBetweenFactor` for the Pose2SLAM problem.
* A `Pose2SLAM` benchmark using all of the above.

To run the benchmark: `swift build -c release -Xswiftc -cross-module-optimization && .build/release/Benchmarks --filter GenericFactorGraph`

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -5,8 +5,8 @@
         "package": "Penguin",
         "repositoryURL": "https://github.com/saeta/penguin.git",
         "state": {
-          "branch": "master",
-          "revision": "3081343f667134e6a7e43701c2c91e433567abea",
+          "branch": "marcrasi/arraybuffer_cast",
+          "revision": "fb901e91cabfdb8ff409020aea928ff4452ed3ab",
           "version": null
         }
       },
--- a/Package.swift
+++ b/Package.swift
@@ -19,7 +19,7 @@ let package = Package(
     // Dependencies declare other packages that this package depends on.
     // .package(url: /* package url */, from: "1.0.0"),
     .package(url: "https://github.com/google/swift-benchmark.git", .branch("master")),
-    .package(url: "https://github.com/saeta/penguin.git", .branch("master")),
+    .package(url: "https://github.com/saeta/penguin.git", .branch("marcrasi/arraybuffer_cast")),
   ],
   targets: [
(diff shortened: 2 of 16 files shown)


ProfFan:
Wow!

ProfFan:
Wild guess: 13x speedup?
```
name                                                                                 time             std        iterations
---------------------------------------------------------------------------------------------------------------------------
Pose2SLAM: NonlinearFactorGraph, Intel, 5 Gauss-Newton steps, 100 CGLS steps         13245027430.0 ns ±   0.00 %          1
Pose2SLAM: GenericFactorGraph, Intel, 10 Gauss-Newton steps, 500 CGLS steps           4477341154.0 ns ±   0.00 %          1
```

marcrasi:
The result on my machine is 1.8s. For comparison:
* The existing SwiftFusion NonlinearFactorGraph: 110s
* GTSAM configured to do a similar thing: 4.1s
* The prototype from ~2 weeks ago: 0.4s.

Why is this 5x slower than the prototype from ~2 weeks ago? I think it's because there is some additional abstraction that the Swift optimizer isn't optimizing away. In principle the optimizer should be able to optimize that all away and get us to 0.4s. We might have to adjust the implementation a bit to make this happen. I'm focusing on getting the APIs and documentation cleaned up before tomorrow, so I might not have time to make it faster by tomorrow, but I do want to get it to 0.4s soon.

dellaert:
Interesting! OK, good plan.

marcrasi:
Adds:
* `VariableAssignments` methods that make it nicer and easier to use.
* `GenericFactorGraph` and `GenericGaussianFactorGraph` with high-level methods that make it easy to build a graph and linearize it.
* `GenericJacobianFactor` for linearizing factors.
* `GenericPriorFactor` and `GenericBetweenFactor` for the Pose2SLAM problem.
* A `Pose2SLAM` benchmark using all of the above.

To run the benchmark: `swift build -c release -Xswiftc -cross-module-optimization && .build/release/Benchmarks --filter GenericFactorGraph`

PR #84, opened 2020-06-13
add documentation about jacobians and specializations

marcrasi:
I realized that it might not be clear how to get the prior and between factors working with Pose3. The `JacobianRows` type parameter is non obvious and not documented. So I added some doc comments about that.

Also I added some specializations that you will need for Pose3SLAM to be fast, and comments about them.

### DIFF
--- a/Sources/SwiftFusion/Inference/FactorsStorage.swift
+++ b/Sources/SwiftFusion/Inference/FactorsStorage.swift
@@ -275,8 +275,18 @@ extension ArrayStorageImplementation where Element: NewGaussianFactor {
   }
 
   /// Returns the results of the factors' linear functions at the given point.
+  // For reasonable performance, this must be specialized for all gaussian factor types that are
+  // used in inner loops.
+  @_specialize(where Self == GaussianFactorArrayStorage<ScalarJacobianFactor<Vector1>>)
+  @_specialize(where Self == GaussianFactorArrayStorage<ScalarJacobianFactor<Vector2>>)
+  @_specialize(where Self == GaussianFactorArrayStorage<ScalarJacobianFactor<Vector3>>)
+  @_specialize(where Self == GaussianFactorArrayStorage<ScalarJacobianFactor<Vector4>>)
+  @_specialize(where Self == GaussianFactorArrayStorage<ScalarJacobianFactor<Vector5>>)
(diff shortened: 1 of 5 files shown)


ProfFan:
@marcrasi Thanks a lot!  I figured out the `Array6<Tuple<Array6, Array6>` part but didn't realize that we need the specializations for performance :) And Pose3SLAM is sort of working! I'll post some results after merging this PR and rebasing the LM WIP.

ProfFan:
@marcrasi Now that we have the solution, how can we access the stored variables? I am thinking of storing the `TypedID`s in `G2ONewFactorGraph`, so we probably need a type-erased container for `TypedID`?

marcrasi:
You want to read the values from a `VariableAssignments`?

A quick and dirty way to do this would be `x[TypedID<Pose2, Int>(0)]`. This works because you know the variable types and ids, but it's a bit dirty because the G2O reader also "knows" the types and ids and so it should be able to give them to you without you having to specify the types.

marcrasi:
Yeah, storing `TypedIDs` in `G2ONewFactorGraph` would be a nice clean solution. I think you can achieve this without type-erasure, by adding a type parameter on `G2ONewFactorGraph`:

```
  public struct G2ONewFactorGraph<Pose: LieGroup> {
    /// Mapping from g2o variable index to pose variable.
    public var poseIDs: [Int: TypedID<Pose, Int>]

    /// The initial guess.
    public var initialGuess = VariableAssignments()

    /// The factor graph representing the measurements.
    public var graph = NewFactorGraph()    

    /// Creates a problem from the given 2D file.
    public init(g2oFile2D: URL) throws where Pose == Pose2 {
      try G2OReader.read2D(file: g2oFile2D) { entry in
        switch entry {
        case .initialGuess(index: let index, pose: let guess):
          poseIDs[index] = initialGuess.store(guess)
        case .measurement(frameIndex: let index1, measuredIndex: let index2, pose: let difference):
          graph.store(NewBetweenFactor2(poseIDs[index1], poseIDs[index2], difference))
        }
      }
    }
  }
```

This also makes the reader itself better because it removes the assumption that all the indices are consecutive!

(I haven't tried to compile this, there may be typos or other problems.)

ProfFan:
@marcrasi Thanks! I got it working, will prepare a PR soon.

marcrasi:
Cool! fyi, I'll be off the internet for the rest of today so sorry if I don't respond to any more questions until tomorrow :)

PR #85, opened 2020-06-15
[WIP] Add LM Pose3Example and G2O 3D Reader

ProfFan:


ProfFan:
One important note here: the error (referring to the CGLS book Bjorck96) should be calculated as `b - Ax`

ProfFan:
After 30 GN steps, `sphere2500` dataset:
image

Cost 185s of runtime, 30 iterations, cross-module optimization enabled (1.3s, 7 iterations with GTSAM). Since we still do not have a noise model, it's not a particularly fair comparison in number of iterations, but the performance difference in time per iteration is still there (6s v.s. 0.18s)

@marcrasi 

Test environment:
macOS 10.15, i7-8700K 4.7GHz

ProfFan:
Add reference to #42

Sources/Benchmarks/Pose3SLAM.swift
```diff
@@ -49,4 +48,33 @@ let pose3SLAM = BenchmarkSuite(name: "Pose3SLAM") { suite in
...
   }
+  
+  var gridDataset =  try! G2OReader.G2ONewFactorGraph(g2oFile3D: try! cachedDataset("pose3example.txt"))
+        // try! G2OReader.G2ONewFactorGraph(g2oFile3D: try! cachedDataset("sphere2500.g2o"))
+  //  check(gridDataset.graph.error(gridDataset.initialGuess), near: 12.99, accuracy: 1e-2)
+  
+  // Uses `NonlinearFactorGraph` on the Intel dataset.
+  // The solvers are configured to run for a constant number of steps.
+  // The nonlinear solver is 5 iterations of Gauss-Newton.
+  // The linear solver is 100 iterations of CGLS.
```
marcrasi, line 59: This comment talks about a different benchmark.

Sources/SwiftFusion/Datasets/G2OReader.swift
```diff
@@ -47,25 +47,42 @@ public enum G2OReader {
   }
 
   /// A G2O problem expressed as a `NewFactorGraph`.
-  public struct G2ONewFactorGraph {
+  public struct G2ONewFactorGraph<Pose: LieGroup> {
     /// The initial guess.
     public var initialGuess = VariableAssignments()
 
+    public var initialGuessId = Array<TypedID<Pose, Int>>()
```
marcrasi, line 54: I suggest naming this something that refers to "variable id" instead of "initial guess" because identifies variables generally not just initial guesses. Like: `poseVariableID`

Sources/SwiftFusion/Datasets/G2OReader.swift
```diff
@@ -47,25 +47,42 @@ public enum G2OReader {
...
           let typedID = initialGuess.store(guess)
           assert(typedID.perTypeID == id)
+          initialGuessId.append(typedID)
         case .measurement(frameIndex: let id1, measuredIndex: let id2, pose: let difference):
           graph.store(NewBetweenFactor2(TypedID(id1), TypedID(id2), difference))
         }
       }
     }
+    
+    /// Creates a problem from the given 2D file.
```
marcrasi, line 73: s/2D/3D/

Tests/SwiftFusionTests/Inference/NewFactorGraphTests.swift
```diff
@@ -46,4 +46,71 @@ class NewFactorGraphTests: XCTestCase {
...
+  /// keys starting from 0. An optional character may be provided, which
+  /// will be stored in the msb of each key (i.e. gtsam.Symbol).
+
+  /// We use aerospace/navlab convention, X forward, Y right, Z down
+  /// First pose will be at (R,0,0)
+  /// ^y   ^ X
+  /// |    |
+  /// z-->xZ--> Y  (z pointing towards viewer, Z pointing away from viewer)
+  /// Vehicle at p0 is looking towards y axis (X-axis points towards world y)
+  func circlePose3(numPoses: Int = 8, radius: Double = 1.0) -> Values {
```
marcrasi, line 61: Could this create a `VariableAssignments` directly so that you don't have to convert it later?

Sources/SwiftFusion/Inference/NewJacobianFactor.swift
```diff
@@ -62,7 +62,7 @@ public struct NewJacobianFactor<
   }
 
   public func errorVector(at x: Variables) -> ErrorVector {
-    return errorVector_linearComponent(x) + error
+    return error - errorVector_linearComponent(x)
```
marcrasi, line 65: I'm pretty confused by the signs.

I see that we need to calculate `b - Ax`, but I'm not sure if `errorVector` should be the method that does this.

Changing `errorVector` to `b - Ax` changes the meaning of linearization (the comment on `NewFactorGraph.linearized` implies that `errorVector` should be `Ax + b`).

Also, changing `errorVector` to `b - Ax` makes it so that `errorVector_linearComponent` isn't actually the linear component of `errorVector` and more.

Maybe we should make a new method called `residual` that calculates `b - Ax`?

Or we could negate the implementations of `errorVector_linearComponent` and `errorVector_linearComponent_adjoint`, and leave `errorVector(at: x) = errorVector_linearComponent(x) + error`. Then we'd also have to update the comment on `NewFactorGraph.linearized` to say:
```
  /// The linear approximation satisfies the approximate equality:
  ///   `self.linearized(at: x).errorVectors(at: dx)` ≈ `self.errorVectors(at: x.moved(along: -1 * dx))`
  /// where the equality is exact when `dx == x.linearizedZero`.
```

Sources/Benchmarks/Pose3SLAM.swift
ProfFan: Done

Sources/SwiftFusion/Datasets/G2OReader.swift
ProfFan: Agreed :) Done.

Tests/SwiftFusionTests/Inference/NewFactorGraphTests.swift
ProfFan: Modified to return `(hexagonId, hexagon)`, but I still feel that Frank may disagree with the current `TypedID` API as we effectively lost control of what the ID will be.

Sources/SwiftFusion/Inference/NewJacobianFactor.swift
ProfFan: Yea I feel a bit confused too :( IMHO, in CGLS the objective is to reach the solution where `Ax=b`, thus by moving Ax to the right we have `0 = b - Ax`(the "error"). In GN we approximate the L2 error function (which should be >=0) `f(x)` by its linearization at `x0`, where `f(dx+x_0)=Ax+b` (Jacobian `A`, bias `b`), and we arrived at the same equation as in CGLS.

I think `error` here is confusing as it is actually the "error of error", as what we linearized to the JacobianFactor IS the error. However, I am not sure if changing this is actually a good idea, as if we want linear factors to be nonlinear compatible we need the protocol conformances.

dellaert: nonlinear error: |h(x)-z|
linear error |Ax-b|

ProfFan:
I merged so we can start using the LM solver right now, and do optimizations later (((

PR #87, opened 2020-06-17
add Pose3 custom Adjoint, to make Pose3SLAM faster

marcrasi:
I changed the Pose3SLAM benchmark to use sphere2500 instead of pose3example because pose3example is so small that I was getting very noisy benchmark results.

Results from the sphere2500 benchmark:
* Before: 108s
* After: 20s

### DIFF
--- a/Sources/Benchmarks/Pose3SLAM.swift
+++ b/Sources/Benchmarks/Pose3SLAM.swift
@@ -48,8 +48,8 @@ let pose3SLAM = BenchmarkSuite(name: "Pose3SLAM") { suite in
     }
   }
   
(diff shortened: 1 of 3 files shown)


ProfFan:
Duh! This is awesome!

marcrasi:
You're wondering if the compiler could inline all the code for the adjoint and then do mathematical simplifications to automatically get something as fast as this implementation?

I haven't tried to do this. It would be interesting to try some day. I expect that the optimizer won't know enough mathematical simplifications to make the code nearly as fast as the custom one, but it would be interesting to see what is missing.

`@inline(__always)` is the attribute that would help with this experiment. (`@inlinable` is for making things inlinable into callers from other modules, but we're calling within the same module, so `@inlinable` has no effect here.)

PR #89, opened 2020-06-17
speed up the new factor graphs

marcrasi:
Adds more specializations, and changes some particularly hot code to use pointers instead of collections.

Makes Pose2SLAM ~4x faster and Pose3SLAM ~3x faster.

Before:
```
swift build -c release -Xswiftc -cross-module-optimization && .build/release/Benchmarks --filter NewFactorGraph
name                                                                    time             std        iterations
--------------------------------------------------------------------------------------------------------------
Pose2SLAM: NewFactorGraph, Intel, 10 Gauss-Newton steps, 500 CGLS steps  1868228204.0 ns ±   0.00 %          1
Pose3SLAM: NewFactorGraph, sphere2500, 30 LM steps, 200 CGLS steps      19669232876.0 ns ±   0.00 %          1
```

After:
```
swift build -c release -Xswiftc -cross-module-optimization && .build/release/Benchmarks --filter NewFactorGraph
name                                                                    time            std        iterations
-------------------------------------------------------------------------------------------------------------
Pose2SLAM: NewFactorGraph, Intel, 10 Gauss-Newton steps, 500 CGLS steps  444424233.0 ns ±   0.00 %          1
Pose3SLAM: NewFactorGraph, sphere2500, 30 LM steps, 200 CGLS steps      6116728990.0 ns ±   0.00 %          1
```

### DIFF
--- a/Sources/SwiftFusion/Inference/NewJacobianFactor.swift
+++ b/Sources/SwiftFusion/Inference/NewJacobianFactor.swift
@@ -66,11 +66,26 @@ public struct NewJacobianFactor<
   }
 
   public func errorVector_linearComponent(_ x: Variables) -> ErrorVector {
-    return ErrorVector(jacobian.lazy.map { $0.dot(x) })
+    // The compiler isn't able to optimize the closure away if we map `jacobian`, but it is able
+    // to optimize the closure away if we map `jacobian`'s `UnsafeBufferPointer`.
+    jacobian.withUnsafeBufferPointer { rows in
+      ErrorVector(rows.lazy.map { $0.dot(x) })
+    }
   }
 
   public func errorVector_linearComponent_adjoint(_ y: ErrorVector) -> Variables {
-    return zip(y.scalars, jacobian).lazy.map(*).reduce(Variables.zero, +)
+    // We use `UnsafeBufferPointer`s to avoid forming collections that can't be optimized away.
+    y.withUnsafeBufferPointer { scalars in
+      jacobian.withUnsafeBufferPointer { rows in
+        // We reduce the range `0..<ErrorVector.dimension` instead of `zip(scalars, rows)`, to
+        // avoid forming collections that can't be optimized away.
+        // TODO: This is not getting unrolled after `ErrorVector` is specialized. Convincing the
+        // optimizer to unroll it might speed things up.
+        (0..<ErrorVector.dimension).reduce(into: Variables.zero) { (result, i) in
+          result += scalars[i] * rows[i]
+        }
+      }
+    }
   }
 
   public typealias Linearization = Self
--- a/Sources/SwiftFusion/Core/EuclideanVectorN.swift
+++ b/Sources/SwiftFusion/Core/EuclideanVectorN.swift
@@ -6,3 +6,31 @@ public protocol EuclideanVectorN: EuclideanVector {
   /// A standard basis of vectors.
   static var standardBasis: [Self] { get }
 }
+
+extension EuclideanVectorN {
+  /// Returns the result of calling `body` on the scalars of `self`.
+  func withUnsafeBufferPointer<R>(
+    _ body: (UnsafeBufferPointer<Double>) throws -> R
+  ) rethrows -> R {
+    return try withUnsafePointer(to: self) { p in
+      try body(
+          UnsafeBufferPointer<Double>(
+              start: UnsafeRawPointer(p)
+                  .assumingMemoryBound(to: Double.self),
+              count: Self.dimension))
+    }
+  }
+
+  /// Returns the result of calling `body` on the scalars of `self`.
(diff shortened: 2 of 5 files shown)


Sources/SwiftFusion/Inference/NewJacobianFactor.swift
```diff
@@ -66,11 +66,25 @@ public struct NewJacobianFactor<
   }
 
   public func errorVector_linearComponent(_ x: Variables) -> ErrorVector {
-    return ErrorVector(jacobian.lazy.map { $0.dot(x) })
+    var r = ErrorVector.zero
```
dabrahams, line 69: This one surprises me, because in theory it writes on `r` twice. When forced to write something out the ugly way or with a raw loop, or especially with code that looks like it shold be inefficient, it's a good idea to put in a comment that explains both what it does, and why you wrote it that way so someone doesn't come along later and say “oh, that's a one-liner” and rewrite it.
```suggestion
    // ErrorVector(jacobian.lazy.map { $0.dot(x) }), optimized
    var r = ErrorVector.zero
```

Sources/SwiftFusion/Inference/NewJacobianFactor.swift
```diff
@@ -66,11 +66,25 @@ public struct NewJacobianFactor<
...
+    jacobian.withUnsafeBufferPointer { jacRows in
+      r.withUnsafeMutableBufferPointer { rScalars in
+        for i in 0..<ErrorVector.dimension {
+          rScalars[i] = jacRows[i].dot(x)
+        }
```
dabrahams, line 74: I started to make all kindsa suggestions here about other approaches and then I realized I was confused about what's being computed here.

----

Would it be good (and optimize well) to write this inner loop as a generic algorithm over collections, so
```suggestion
        rScalars.inplaceDot(jacRows)
```
?

Similarly, once done, you might try:

```
r.inplaceDot(jacobian)
``` 

or even

```
jacobian

File: Sources/SwiftFusion/Inference/NewJacobianFactor.swift [hunk 3]
```diff
@@ -66,11 +66,25 @@ public struct NewJacobianFactor<
...
+    }
+    return r
   }
 
   public func errorVector_linearComponent_adjoint(_ y: ErrorVector) -> Variables {
-    return zip(y.scalars, jacobian).lazy.map(*).reduce(Variables.zero, +)
+    y.withUnsafeBufferPointer { yScalars in
+      jacobian.withUnsafeBufferPointer { jacRows in
+        (0..<ErrorVector.dimension).reduce(into: Variables.zero) { (result, i) in
+          result += yScalars[i] * jacRows[i]
```
dabrahams, line 84: Can't this be expressed as some kind of `dot`?

dabrahams:
Also, after using the new dispatch stuff it would be good to check to see if all of these hand-optimizations are still needed.  Sucks, right?

Sources/SwiftFusion/Inference/NewJacobianFactor.swift
marcrasi: I figured out a way to get the oneliner with `map` to optimize well. I'll upload the commit soon.

Sources/SwiftFusion/Inference/NewJacobianFactor.swift
marcrasi: May not be relevant once I bring the oneliner back.

Sources/SwiftFusion/Inference/NewJacobianFactor.swift
marcrasi: If we had some way of accessing the columns of the jacobian, then it could be a dot:
```
Variables(jacobian.columns.lazy.map { col in col.dot(y) })
```

But writing the `.columns` collection seems like complicated overkill and I expect it also wouldn't optimize well.

We can't simply switch `jacobian` to column-major because `errorVector_linearComponent` needs to access the rows.

dabrahams: We need .columns for usability anyway, and I don't know why you think it wouldn't optimize well; maybe it would, maybe not, but it's worth doing.

marcrasi: I would like to defer `.columns` until later because it's pretty non-trivial to "calculate" the type of `.columns`.

The rows are fixed-size arrays of tuple of vectors: types like `Array6<Tuple2<Vector3, Vector3>>`. I believe there's no way to "transpose" these types to the column types like `Tuple2<Array3<Vector6>, Array3<Vector6>>`.

The right solution is probably to define fixed-size matrices that know their "rows" type and their "columns" type. But there is a lot of fixed-size matrix design that we would need to do that is out of scope of this PR.

PR #92, opened 2020-06-18
Migrate to new ArrayBuffer

marcrasi:
Migrates everything to the new `ArrayBuffer` interface from https://github.com/saeta/penguin/pull/82.

The boilerplate in `ValuesStorage` and `FactorsStorage` feels simpler now, which is good!

Benchmarks before:
```
swift build -c release -Xswiftc -cross-module-optimization && .build/release/SwiftFusionBenchmarks --filter NewFactorGraph
name                                                                   time           std        iterations
-----------------------------------------------------------------------------------------------------------
Pose2SLAM.NewFactorGraph, Intel, 10 Gauss-Newton steps, 500 CGLS steps  456.792364 ms ±   0.00 %          1
Pose3SLAM.NewFactorGraph, sphere2500, 30 LM steps, 200 CGLS steps      5922.074432 ms ±   0.00 %          1
```

Benchmarks after:
```
swift build -c release -Xswiftc -cross-module-optimization && .build/release/SwiftFusionBenchmarks --filter NewFactorGraph
name                                                                   time           std        iterations
-----------------------------------------------------------------------------------------------------------
Pose2SLAM.NewFactorGraph, Intel, 10 Gauss-Newton steps, 500 CGLS steps  446.596137 ms ±   0.00 %          1
Pose3SLAM.NewFactorGraph, sphere2500, 30 LM steps, 200 CGLS steps      4950.754052 ms ±   0.00 %          1
```

### DIFF
--- a/Sources/SwiftFusion/Inference/PenguinExtensions.swift
+++ b/Sources/SwiftFusion/Inference/PenguinExtensions.swift
@@ -1,0 +1,55 @@
+// TODO: Move all these to Penguin.
+
+import PenguinStructures
+
+extension AnyArrayBuffer {
+  /// Appends `x`, returning the index of the appended element.
+  ///
+  /// - Requires: `self.elementType == T.self`.
+  mutating func unsafelyAppend<T>(_ x: T) -> Int {
+    unsafelyMutate(assumingElementType: Type<T>()) { $0.append(x) }
+  }
+}
+
+/// An `AnyArrayBuffer` dispatcher for elements that are not statically known to support any
+/// operations.
+// We can't use `AnyObject` for this because the bitcast at [1] fails with "Swift runtime failure:
+// Can't unsafeBitCast between types of different sizes" when trying to cast an `AnyArrayBuffer`
+// of some other dispatch to an `AnyArrayBuffer<AnyObject>`.
+//
+// [1] https://github.com/saeta/penguin/blob/47980eb5630ebfd7bc2a88a4a7a60a050dd01b0a/Sources/PenguinStructures/AnyArrayBuffer.swift#L64
+class AnyElementDispatch {}
+
+/// A type-erased array that is not statically known to support any operations.
+typealias AnyElementArrayBuffer = AnyArrayBuffer<AnyElementDispatch>
+
+extension AnyArrayBuffer where Dispatch == AnyElementDispatch {
+  /// Creates an instance from a typed buffer of `Element`
+  init<Element>(_ src: ArrayBuffer<Element>) {
+    self.init(
+      storage: src.storage,
+      dispatch: AnyElementDispatch.self)
+  }
+}
+
+extension ArrayBuffer {
+  /// Creates an instance with the given `count`, and capacity at least
+  /// `minimumCapacity`, and elements initialized by `initializeElements`,
+  /// which is passed the address of the (uninitialized) first element.
+  ///
+  /// - Requires: `initializeElements` initializes exactly `count` contiguous
+  ///   elements starting with the address it is passed.
+  public init(
+    count: Int,
+    minimumCapacity: Int = 0,
+    initializeElements:
+      (_ uninitializedElements: UnsafeMutablePointer<Element>) -> Void
+  ) {
+    // Work around the fact that `ArrayBuffer.init(storage:)` doesn't exist. We can just set
+    // `storage` directly without calling a different initializer first when we move this
+    // implementation to penguin.
+    self.init(minimumCapacity: 0)
+    self.storage = .init(
+      count: count, minimumCapacity: minimumCapacity, initializeElements: initializeElements)
+  }
+}
--- a/Sources/SwiftFusion/Inference/ValuesStorage.swift
+++ b/Sources/SwiftFusion/Inference/ValuesStorage.swift
@@ -17,368 +17,371 @@
 
 import PenguinStructures
 
-// MARK: - Storage for contiguous arrays of `Differentiable` values.
+// MARK: - Algorithms on arrays of `Differentiable` values.
 
-/// Contiguous storage of homogeneous `Differentiable` values of statically unknown type.
-class AnyDifferentiableStorage: AnyArrayStorage {
-  typealias DifferentiableImplementation = AnyDifferentiableStorageImplementation
-  var differentiableImplementation: DifferentiableImplementation {
-    fatalError("implement me!")
-  }
-
-  /// Identifies the values' `TangentVector`.
-  var tangentIdentifier: ObjectIdentifier {
-    differentiableImplementation.tangentIdentifier_
-  }
-
-  /// The values' zero `TangentVector`s.
-  var zeroTangent: AnyArrayBuffer<AnyVectorStorage> {
-    differentiableImplementation.zeroTangent_
-  }
-  
-  /// Moves `self` along the vector starting at `directionsStart`.
-  ///
-  /// Precondition: `directionsStart` points to memory with at least `count` initialized
-  /// `Element.TangentVector`s where `Element` is the element type of `self`.
-  final func move(along directionsStart: UnsafeRawPointer) {
-    differentiableImplementation.move_(along: directionsStart)
-  }
-}
-
-extension AnyArrayBuffer where Storage: AnyDifferentiableStorage {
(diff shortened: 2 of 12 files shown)


Sources/SwiftFusion/Inference/PenguinExtensions.swift
```diff
@@ -0,0 +1,55 @@
...
+  }
+}
+
+/// An `AnyArrayBuffer` dispatcher for elements that are not statically known to support any
+/// operations.
+// We can't use `AnyObject` for this because the bitcast at [1] fails with "Swift runtime failure:
+// Can't unsafeBitCast between types of different sizes" when trying to cast an `AnyArrayBuffer`
+// of some other dispatch to an `AnyArrayBuffer<AnyObject>`.
+//
+// [1] https://github.com/saeta/penguin/blob/47980eb5630ebfd7bc2a88a4a7a60a050dd01b0a/Sources/PenguinStructures/AnyArrayBuffer.swift#L64
```
dabrahams, line 20: Something's not right here.  The `MemoryLayout<AnyObject.Type>.size` is the same as the size of any class type.  You should put some print statements in there to see what it thinks the types and sizes are.

Some weirdness in this area:
https://bugs.swift.org/browse/SR-13055

Sources/SwiftFusion/Inference/PenguinExtensions.swift
```diff
@@ -0,0 +1,55 @@
...
+class AnyObjectDispatch {}
```
dabrahams, line 21:
```suggestion
class AnyElementDispatch {}
```

Sources/SwiftFusion/Inference/ValuesStorage.swift
```diff
@@ -17,368 +17,369 @@
...
-  var tangentIdentifier: ObjectIdentifier {
-    storage.tangentIdentifier
+// MARK: - Algorithms on arrays of `EuclideanVectorN` values.
+
+extension ArrayStorage where Element: EuclideanVectorN {
+  /// Adds `others` to `self`.
+  ///
+  /// - Requires: `others.count >= count`.
+  func add(_ others: ArrayBuffer<Element>) {
+    withUnsafeMutableBufferPointer { vs in
```
dabrahams, line 51: I don't love the code pattern repetition in these three methods; please refactor if you need to do it again.

Sources/SwiftFusion/Inference/ValuesStorage.swift
```diff
@@ -17,368 +17,369 @@
...
+      dispatch: VectorArrayDispatch_<Element>.self)
+  }
+}
+
+extension AnyArrayBuffer where Dispatch: VectorArrayDispatch {
   /// Adds `others` to `self`.
   ///
-  /// Precondition: `others.count >= count`.
-  mutating func add<Others: Collection>(_ others: Others) where Others.Element == Element {
+  /// - Requires: `others.count >= count`.
```
dabrahams, line 357: 🤔

Sources/SwiftFusion/Inference/ValuesStorage.swift
```diff
@@ -17,368 +17,369 @@
...
   }
 
   /// Returns the dot product of `self` with `others`.
   ///
   /// This is the sum of the dot products of corresponding elements.
   ///
-  /// Precondition: `others.count >= count`.
-  func dot<Others: Collection>(_ others: Others) -> Double where Others.Element == Element {
-    storage.dot(others)
+  /// - Requires: `others.count >= count`.
```
dabrahams, line 373: 🤔

Sources/SwiftFusion/Inference/ValuesStorage.swift
```diff
@@ -17,368 +17,369 @@
...
   ///
-  /// Precondition: `direction` has at least `count` elements of type `Element.TangentVector`,
-  /// where `Element` is the type of `self`.
-  mutating func move<DirectionStorage>(along directions: AnyArrayBuffer<DirectionStorage>) {
-    ensureUniqueStorage()
-    directions.withUnsafeRawPointerToElements { directionsStart in
-      storage.move(along: directionsStart)
+  /// This is the sum of the dot products of corresponding elements.
+  ///
+  /// - Requires: `others.count >= count`.
```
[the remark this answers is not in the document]
marcrasi: will change and add an assert

Sources/SwiftFusion/Inference/ValuesStorage.swift
```diff
@@ -17,368 +17,369 @@
...
 
-  /// Moves `self` along the vector starting at `directionsStart`.
+/// An `AnyArrayBuffer` dispatcher that provides algorithm implementations for `Differentiable`
+/// elements.
+class DifferentiableArrayDispatch: AnyObjectDispatch {
+  /// The `TangentVector` type of the elements.
+  class var tangentVectorType: Any.Type {
+    fatalError("implement as in DifferentiableArrayDispatch_")
+  }
+
```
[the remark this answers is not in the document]
marcrasi: It is used in `VariableAssignments`'s `tangentVectorZeros` to get the dictionary key for the dictionary of tangent vectors. (When we switch to the new TypedID thing that lets you have custom keys we won't need this any more because then the key will be the type of the original variable that it comes from.)

I'll add a test for now.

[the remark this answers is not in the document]
marcrasi: actually, there already is a test

Sources/SwiftFusion/Inference/ValuesStorage.swift
```diff
@@ -17,368 +17,369 @@
...
   }
 
-  /// For each variable, a Jacobian factor that scales it by `scalar`.
-  final func jacobians(scalar: Double)
-    -> (ObjectIdentifier, AnyArrayBuffer<AnyGaussianFactorStorage>)
+  /// For each element of `storage`, a Jacobian that scales it by `scalar`.
+  ///
+  /// - Requires: `storage` is the address of an `ArrayStorage` whose `Element` has a
+  ///   subclass-specific `EuclideanVectorN` type.
+  class func jacobians(_ storage: UnsafeRawPointer, scalar: Double)
```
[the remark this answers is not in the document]
marcrasi: It returns an array of jacobians, and the content of each jacobian is a scalar. So I thought that `jacobians(_:scalar:)` is a good name.

See https://github.com/borglab/SwiftFusion/pull/83 for the motivation. That PR description has some guess at a better, less hardcoded way, of achieving the same thing. We should discuss this some time.

Sources/SwiftFusion/Inference/PenguinExtensions.swift
marcrasi:
```
  /// Creates an instance containing the same elements as `src`.
  ///
  /// - Requires: `src.dispatch is Dispatch.Type`.
  public init<OtherDispatch>(unsafelyCasting src: AnyArrayBuffer<OtherDispatch>)
  {
    assert(src.dispatch is Dispatch.Type)
    print("Dispatch.Type: \(Dispatch.Type.self) \(MemoryLayout<Dispatch.Type>.size)")
    print("OtherDispatch.Type: \(OtherDispatch.Type.self) \(MemoryLayout<OtherDispatch.Type>.size)")
    self.storage = src.storage
    self.dispatch = unsafeBitCast(src.dispatch, to: Dispatch.Type.self)
  }
```

Result with `-O`:
```
Dispatch.Type: AnyObject.Protocol 8
OtherDispatch.Type: DifferentiableArrayDispatch.Type 8
... execution proceeds normally ...
```

Result with `-O -cross-module-optimization`:
```
Dispatch.Type: AnyObject.Protocol 0
OtherDispatch.Type: DifferentiableArrayDispatch.Type 8
Illegal instruction
```

Is this a bug in swiftc? I don't know enough about what is supposed to happen to tell.

dabrahams: Yes, most definitely a bug when the size of `AnyObject.Protocol` changes based on optimization.

marcrasi: filed https://bugs.swift.org/browse/SR-13061

PR #93, opened 2020-06-18
Some example graphs [WIP]

dellaert:
Adding a tracking and switching model


dellaert:
As well, as, with newest commit, a discrete factor is not a GaussianFactor, I guess:

/Users/dellaert/git/SwiftFusion/Tests/SwiftFusionTests/Inference/Scratch.swift:151:5: Instance method 'store' requires that 'DiscreteTransitionFactor' conform to 'NewGaussianFactor'
/Users/dellaert/git/SwiftFusion/Tests/SwiftFusionTests/Inference/Scratch.swift:151:5: Where 'T' = 'DiscreteTransitionFactor', 'NewLinearizableFactor' = 'NewGaussianFactor' (SwiftFusion.NewFactorGraph)

Tests/SwiftFusionTests/Inference/Scratch.swift
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  public func errorVector(at x: Variables) -> Pose.TangentVector {
+    return errorVector(x.head, x.tail.head, x.tail.tail.head)
+  }
+
+  public typealias Linearization = NewJacobianFactor<JacobianRows, ErrorVector>
+  public func linearized(at x: Variables) -> Linearization {
+    Linearization(linearizing: errorVector, at: x, edges: edges)
+  }
```
marcrasi, line 78: The linearization of this factor has 2 inputs because the discrete input to the nonlinear factor has no corresponding input in the linearization.

So you need to help `init(linearizing:at:edges)` figure out what the relevant inputs are:
* Partially apply `errorVector` to the discrete input before passing it to `linearizing:`
* Construct the 2-tuple of differentiable variable ids, and pass it to `edges:`

I'll reply with some working code in a moment.

marcrasi:
> As well, as, with newest commit, a discrete factor is not a GaussianFactor, I guess:
> /Users/dellaert/git/SwiftFusion/Tests/SwiftFusionTests/Inference/Scratch.swift:151:5: Instance method 'store' requires that 'DiscreteTransitionFactor' conform to 'NewGaussianFactor'
/Users/dellaert/git/SwiftFusion/Tests/SwiftFusionTests/Inference/Scratch.swift:151:5: Where 'T' = 'DiscreteTransitionFactor', 'NewLinearizableFactor' = 'NewGaussianFactor' (SwiftFusion.NewFactorGraph)

That error message is a bit misleading. It's okay if the discrete factor is a `NewFactor` -- it doesn't have to be gaussian. You should conform it to `NewFactor` and then that should work.

Tests/SwiftFusionTests/Inference/Scratch.swift
marcrasi: Here is the code:
```
  public func linearized(at x: Variables) -> Linearization {
    let (start, label, end) = (x.head, x.tail.head, x.tail.tail.head)
    let (startEdge, endEdge) = (edges.head, edges.tail.tail.head)
    let differentiableVariables = Tuple2(start, end)
    let differentiableEdges = Tuple2(startEdge, endEdge)
    return Linearization(
      linearizing: { differentiableVariables in
        let (start, end) = (differentiableVariables.head, differentiableVariables.tail.head)
        return errorVector(start, label, end)
      },
      at: differentiableVariables,
      edges: differentiableEdges
    )
  }
```

dellaert:
> That error message is a bit misleading. It's okay if the discrete factor is a `NewFactor` -- it doesn't have to be gaussian. You should conform it to `NewFactor` and then that should work.

OK, that's fixed

Tests/SwiftFusionTests/Inference/Scratch.swift
marcrasi: Nothing very interesting is happening inside that `linearized` method. I'll try to eliminate it in my boilerplate-elimination work this week.

dellaert:
Things compile. Things still left to do:

- [ ] add prior on q1
- [ ] make optimize to correct values (it goes berserk now)
- [ ] add Metropolis algorithm

dellaert:
So @marcrasi @ProfFan @dabrahams , I am seeing similar non-deterministic behavior when optimizing tracking graph. Sometimes it's just fine, other times:

```
[LM INNER] reached the target precision, exiting
Pose2(rot: Rot2(theta: 0.0), t: Vector2(x: 9.066361829818925e+49, y: 0.0))
Pose2(rot: Rot2(theta: 0.0), t: Vector2(x: 1.1906336586214911e+50, y: 0.0))
Pose2(rot: Rot2(theta: 0.0), t: Vector2(x: 1.4160743222808171e+50, y: 0.0))
```

As a hint, the switching example *never* converges to the right values, and is also non-deterministic.

marcrasi:
I debugged the convergence and pushed a commit that fixes it. The fixes were:
* LM was calling CGLS with precision 0, which caused numerical errors to get really big when CGLS tried to optimize things beyond a sensible precision. I changed the precision to 1e-10. (Maybe it should be configurable).
* LM was comparing nonlinear error with linearized error to calculate "model fidelity". LM was using `graph.errors(at: x)` to calculate nonlinear error. But `graph.errors(at: x)` includes errors from discrete factors that do not appear in the linearized error, so it's not comparable with the linearized error. I added `graph.linearizableErrors(at: x)` that computes nonlinear errors for all the linearizable factors.

ProfFan:
@marcrasi Should we cherry-pick this fix to a new PR and merge to master? I think we are running on fast-track now and it's good to merge fast so we won't deal with a ton of conflicts later.

marcrasi:
Yeah, sounds good, I'll create a PR!

dellaert:
> I debugged the convergence and pushed a commit that fixes it. The fixes were:
> 
> * LM was calling CGLS with precision 0, which caused numerical errors to get really big when CGLS tried to optimize things beyond a sensible precision. I changed the precision to 1e-10. (Maybe it should be configurable).
> * LM was comparing nonlinear error with linearized error to calculate "model fidelity". LM was using `graph.errors(at: x)` to calculate nonlinear error. But `graph.errors(at: x)` includes errors from discrete factors that do not appear in the linearized error, so it's not comparable with the linearized error. I added `graph.linearizableErrors(at: x)` that computes nonlinear errors for all the linearizable factors.

confirming it works for me :-) I'll review the PR with cherry-pick.

dellaert:
@marcrasi stuck as there is no mutating subscript operator in VariableAssignments. See commented out line in `flipAndOptimize`.
I tried a bit to extend but got lost :-)

ProfFan:
@dellaert I think there is one:
```swift
  /// Accesses the stored value with the given ID.
  public subscript<T>(id: TypedID<T, Int>) -> T {
    _read {
      yield storage[ObjectIdentifier(T.self), default: Self.noSuchType]
        .withUnsafeRawPointerToElements { p in
          p.assumingMemoryBound(to: T.self).advanced(by: id.perTypeID).pointee
        }
    }
    _modify {
      defer { _fixLifetime(self) }
      yield &storage[ObjectIdentifier(T.self), default: Self.noSuchType]
        .withUnsafeMutableRawPointerToElements { p in
          p.assumingMemoryBound(to: T.self).advanced(by: id.perTypeID)
        }
        .pointee
    }
  }
```

Demo:
```
  func testVariable() {
    var x = VariableAssignments()
    let a = x.store(Pose2())
    x[a] = Pose2(0,0,11)
    XCTAssertEqual(x[a], Pose2(0,0,11))
  }
```

dellaert:
@ProfFan not in master (which I merged into my branch).
seems to have been removed by @marcrasi  [here](https://github.com/borglab/swiftfusion/commit/3aaf3c279f75b78229e4216b168b0c8ab2c657fd#diff-4de13d76041b7f9f25358b96c04c3c3eL89-L104)

marcrasi:
Yeah, sorry, I removed it while refactoring some stuff without realizing that it would be important soon. Here's a PR that adds it back: https://github.com/borglab/SwiftFusion/pull/98

dellaert:
@marcrasi @ProfFan I made an MCMC example for the bee tracking, not quite complete yet, but the optimization in the MCMC loop seems to fail if there is a label == 1. I always start with zero poses (following @marcrasi 's lead) but the outcome, which should be symmetric to label 2, just fails.I add some output below:

```
[LM INNER] reached the target precision, exiting
0
0
Pose2(rot: Rot2(theta: 0.0), t: Vector2(x: 9.99999999999354, y: 0.0))
Pose2(rot: Rot2(theta: 0.0), t: Vector2(x: 10.999999999990616, y: 0.0))
Pose2(rot: Rot2(theta: 0.0), t: Vector2(x: 11.999999999989308, y: 0.0))
[LM INNER] reached the target precision, exiting
0
2
Pose2(rot: Rot2(theta: 0.03579700819516731), t: Vector2(x: 10.000940203973764, y: -0.01965726611788173))
Pose2(rot: Rot2(theta: 0.14660357826670667), t: Vector2(x: 11.000693389723478, y: -0.023146205232906156))
Pose2(rot: Rot2(theta: -0.31939846428386093), t: Vector2(x: 11.985194415302464, y: 0.06053993955004976))
[LM INNER] reached the target precision, exiting
0
0
Pose2(rot: Rot2(theta: 0.0), t: Vector2(x: 9.999999999993538, y: 0.0))
Pose2(rot: Rot2(theta: 0.0), t: Vector2(x: 10.999999999990616, y: 0.0))
Pose2(rot: Rot2(theta: 0.0), t: Vector2(x: 11.999999999989306, y: 0.0))
[LM INNER] reached the target precision, exiting
0
0
Pose2(rot: Rot2(theta: 0.0), t: Vector2(x: 9.99999999999354, y: 0.0))
Pose2(rot: Rot2(theta: 0.0), t: Vector2(x: 10.999999999990616, y: 0.0))
Pose2(rot: Rot2(theta: 0.0), t: Vector2(x: 11.999999999989308, y: 0.0))
[LM INNER] reached the target precision, exiting
0
1
Pose2(rot: Rot2(theta: -0.03579700819516694), t: Vector2(x: 10.000940203973766, y: 0.019657266117881947))
Pose2(rot: Rot2(theta: -0.14660357826670678), t: Vector2(x: 11.000693389723482, y: 0.023146205232907138))
Pose2(rot: Rot2(theta: 0.31939846428386093), t: Vector2(x: 11.98519441530246, y: -0.06053993955004991))
Fatal error: 'try!' expression unexpectedly raised an error: SwiftFusion.LevenbergMarquardtError(message: "maximum lambda reached, giving up"): file /Users/dellaert/git/SwiftFusion/Tests/SwiftFusionTests/Inference/Scratch.swift, line 244
```

I have no idea what's going on :-)

marcrasi:
I investigated the optimization a bit and partially understood what's going on.

tldr: LM gets close to the solution. Then it fails to make more progress towards the solution and it interprets this failure as an error.

We can work around the problem by ignoring the error from the optimizer. I pushed a commit that does this. It's getting closeish to the solution despite the error.

I haven't figured out why the optimizer is behaving badly near the solution but here are some guesses about possibilities:
* We're asking LM for unreasonably high precision and we should just relax the precision and accept the solution that it's getting.
* The CGLS implementation is breaking near the solution because of roundoff errors. (Preconditioning might help? I read about this a bit and I don't think preconditioning's main purpose is to help with roundoff errors but it might help a bit as a side effect. I'm supposed to work on a preconditioner this week, so we'll see if this helps soon :))
* The linearization implementation is breaking near the solution because of roundoff errors.

ProfFan:
@marcrasi Yeah that is intentional (probably needs to be documented)... I designed the LM to throw if it fails to reach the required precision when reaching max lambda, so the user knows that the optimization has failed.

dellaert:
Thanks both for looking into this. However, one thing I find puzzling is that it happens for label 1 but not for label 2, which are symmetric. Just wondering whether that points to something in pose2 that is not quite right.

Issue #88, opened 2020-06-17
Pose3SLAM benchmark nondeterministic


marcrasi:
I noticed that the Pose3SLAM benchmark gives me different final losses every time that I run it. As far as I know, all the computations should be deterministic.

We should get to the bottom of this because it might indicate a bug. For example, the generic factor graph datastructure uses unsafe memory operations, and we might be reading uninitialized memory somewhere.

ProfFan:
I agree on that, and I observed that for some difficult datasets, the nonlinear loss sometimes go to `nan` which indicates possibly a division-by-zero or something. I suspect a impl error in our Pose3 but I am not super sure. Is there a way to raise an exception when we have `nan`?

marcrasi:
> Is there a way to raise an exception when we have `nan`?

I don't think there's an automatic way to do this.

Manually adding checks to the initializers of all our math types might be a good approach though. It seems like this would catch problems close to where they occur. The check should probably use a function that we define that is controlled by an `#if` flag -- `assert` isn't suitable because it gets disabled in `-O` mode and I thing we'll want to be able to debug nans in `-O` mode.

PR #95, opened 2020-06-23
Feature/mcmc

dellaert:
Prototyped a generic MCMC sampler inspired by tfp.

There is a function sampleChain that takes a TransitionKernel (a protocol), and one concrete implementation which is a random walk Metropolis chain. In the other branch I verified it is generic, but I'll first get y'alls review on the design here. Note I am trying to deliberately stick to the tfp interface, so I expect there might be some design choices there that are not "swifty".


dellaert:
This is just a short-term crutch: it would be nice to be ab le to have complete interop between tfp and swiftFusion, so we can use their fancy samplers, and we could run their models with our samplers. But I have no idea whether this interop can happen in both directions. I do know that tfp

dellaert:
BTW I might not understand import etc. The tests run fine on my mac but the build seems to fail.

Sources/SwiftFusion/MCMC/sample.swift
```diff
@@ -0,0 +1,39 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/// Implements Markov chain Monte Carlo via repeated TransitionKernel steps
+/// Inspired by tfp [sample_chain](https://www.tensorflow.org/probability/api_docs/python/tfp/mcmc/sample_chain)
+public func sampleChain<State, Kernel:TransitionKernel>(_ num_results:Int,
+                                                        _ init_state:State,
+                                                        _ kernel:Kernel,
+                                                        _ num_burnin_steps:Int) -> Array<State>
+where Kernel.State==State {
```
ProfFan, line 21: This should not be necessary along with the previous `<State,` parameter as you can always use `Kernel.State` directly.

Tests/SwiftFusionTests/MCMC/MCMCTests.swift
```diff
@@ -0,0 +1,43 @@
...
+//
+
+import TensorFlow
+import SwiftFusion
+
+import XCTest
+
+class MCMCTests: XCTestCase {
+  
+  override func setUpWithError() throws {
```
ProfFan, line 15: Can be deleted if not used.

Tests/SwiftFusionTests/MCMC/MCMCTests.swift
```diff
@@ -0,0 +1,43 @@
...
+    // Put setup code here. This method is called before the invocation of each test method in the class.
+  }
+  
+  override func tearDownWithError() throws {
```
ProfFan, line 19: same

Tests/SwiftFusionTests/MCMC/MCMCTests.swift
```diff
@@ -0,0 +1,43 @@
...
+    // Put teardown code here. This method is called after the invocation of each test method in the class.
+  }
+  
+  /// Sampling from the Standard Normal Distribution.
+  /// Inspired by testRWM1DUniform from tfp
+  func testRWM1DUniform() throws {
```
ProfFan, line 25: Maybe add some explanation? There are no docs on the TFP side either, but we should document this.

Sources/SwiftFusion/MCMC/sample.swift
dellaert: Good point! Done

Tests/SwiftFusionTests/MCMC/MCMCTests.swift
dellaert: done

Tests/SwiftFusionTests/MCMC/MCMCTests.swift
dellaert: done

Tests/SwiftFusionTests/MCMC/MCMCTests.swift
dellaert: Good point. Done. Let me know if more clear now.

dellaert:
Test failure might be related to chordal, @ProfFan ?

Sources/SwiftFusion/MCMC/TransitionKernel.swift
```diff
@@ -0,0 +1,27 @@
...
+public protocol TransitionKernel {
+  associatedtype State
+  associatedtype Results
+  
+  /// Takes one step of the TransitionKernel.
+  func one_step(_ current_state: State, _ previous_kernel_results: Results) -> (State, Results)
+  
+  /// Returns an object with the same type as returned by one_step(...)[1].
+  func bootstrap_results(_ init_state: State) -> Results
+}
```
marcrasi, line 26: As a direct translation of the Python API, this looks right to me. If you want to go this direction, then that sounds good to me!

---

The "Swifty" way can make some things a lot nicer so I'll describe a "Swifty" idea here to tempt you :)

The `TransitionKernel` looks like an infinite [Sequence](https://developer.apple.com/documentation/swift/sequence). So there's no need to define a `protocol TransitionKernel` -- you can conform things to `Sequence` instead.

For example, `RandomWalkMetropolis` would be something like:

```swift
struct RandomWalkMetropolis<State>: Sequence, IteratorProtocol {

  var previousLogProb: Double
  var currentState: State

  let target_log_prob_fn: (State) -> Double
  let new_state_fn : (State)->State

  init(_ target_log_prob_fn: @escaping (State) -> Double, _ new_state_fn : @escaping (State)->State, _ initialState: State) {
    ...
  }

  mutating func next() -> State {
    let result = currentState

    // update `currentState` and `previousLogProb` according to metropolis rules
    ...

    return result
  }
}
```

This is nice because it allows us to use all the `Sequence` algorithms. For example, `sampleChain` becomes something like:

```swift
Array(sequence.dropFirst(num_burnin_steps).prefix(num_results))
```

Sources/SwiftFusion/MCMC/RandomWalkMetropolis.swift
```diff
@@ -0,0 +1,59 @@
...
+    let new_state = new_state_fn(current_state)
+    let new_log_prob = target_log_prob_fn(new_state)
+    
+    // Calculate log of acceptance ratio p(x')/p(x) = log p(x') - log p(x)
+    let current_log_prob = previous_kernel_results
+    let log_accept_ratio = new_log_prob - current_log_prob
+    
+    // If p(x')/log p(x) >= 1 , i.e., log_accept_ratio >= 0, we always accept
+    // otherwise we accept randomly with probability p(x')/log p(x).
+    // We do this by randomly sampling u from [0,1], and comparing log(u) with log_accept_ratio.
```
marcrasi, line 45: `p(x')/log p(x)` should be `p(x')/p(x)` in two places, I think

marcrasi:
> Test failure might be related to chordal, @ProfFan ?

I think I've noticed these tests being flaky recently. Usually asking github to rerun it fixes it for me. (Though it would also be good to deflake them :))

ProfFan:
> > Test failure might be related to chordal, @ProfFan ?
> 
> I think I've noticed these tests being flaky recently. Usually asking github to rerun it fixes it for me. (Though it would also be good to deflake them :))

Yeah we should at some point try to dig into and fix them (#88). Also need to check for nans in the Pose3SLAM benchmark, as I think even with too big deltas we should not have NaN in the Lie operations.

Sources/SwiftFusion/MCMC/TransitionKernel.swift
dellaert: That is nice, lazy haskell style :-)

I created an issue so we can revisit, chat about larger issue of tfp interop in general.

PR #102, opened 2020-06-25
deflake some tests

marcrasi:
* Sometimes the randomized Rot3 tests fail to be within the accepted accuracy. I expanded the accepted range.
* Sometimes the randomized chordal PoseSLAM test fails because it gets stuck in a local minimum (I think?). I added retries.


marcrasi:
> It may be good to add one additional check on the percentage of randomized test failed, though.

The current behavior of trying 10 times and exiting on success is already sort of equivalent to checking percentage failure: if the optimization fails 90% of the time, then the test fails 0.9^10 = 34% of the time.

An explicit check would allow us to fine tune the false positive / false negative tradeoff but the current behavior seems like a reasonable tradeoff to me.

PR #104, opened 2020-06-26
reduce linearization boilerplate

marcrasi:
Eliminates the `JacobianRows` generic parameter and `linearized` method boilerplate from factor definitions. See `BetweenFactor.swift` for an example.

**How `linearized` was eliminated:**

For factors whose variables are all differentiable, `linearized` is unnecessary because can linearize the factor by differentiating a `@differentiable` `errorVector` method

So I changed `LinearizedFactor` to require that all its variables are differentiable.

To allow for factors with a mix of discrete and differentiable variables, I added a new `VectorFactor` in the factor refinement hierarchy. This factor has a `linearizableComponent` method for "partially applying" the factor to its nondifferentiable variables so that you can linearize with respect to all the other variables using `LinearizedFactor`.

**How `JacobianRows` was eliminated:**

I made `VectorFactorArrayDispatch` responsible for deciding the jacobian type.

### DIFF
--- a/Examples/Pose3SLAMG2O/main.swift
+++ b/Examples/Pose3SLAMG2O/main.swift
@@ -133,7 +133,7 @@ func main() {
   var val = problem.initialGuess
   var graph = problem.graph
   
-  graph.store(PriorFactor3(TypedID(0), Pose3(Rot3.fromTangent(Vector3.zero), Vector3.zero)))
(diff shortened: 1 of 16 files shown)


marcrasi:
Closing because it's all in https://github.com/borglab/SwiftFusion/pull/113.

Issue #105, opened 2020-06-26
Subscripted View for Fixed-Sized Matrices and Vectors


ProfFan:
When implementing the chordal initialization I found out that it is almost essential that we have:
1. a mutable, subscripted view into the fixed-size matrices;
2. methods like `MatrixN.zero` and `MatrixN.identity`.

Otherwise I end up writing code like this:
```
let M9: Jacobian9x9_2 = Array9([
        Tuple2(Vector9(-1,0,0,0,0,0,0,0,0), Vector9(Rij.s00,Rij.s01,Rij.s02,0,0,0,0,0,0)),
        Tuple2(Vector9(0,-1,0,0,0,0,0,0,0), Vector9(Rij.s10,Rij.s11,Rij.s12,0,0,0,0,0,0)),
        Tuple2(Vector9(0,0,-1,0,0,0,0,0,0), Vector9(Rij.s20,Rij.s21,Rij.s22,0,0,0,0,0,0)),
        Tuple2(Vector9(0,0,0,-1,0,0,0,0,0), Vector9(0,0,0,Rij.s00,Rij.s01,Rij.s02,0,0,0)),
        Tuple2(Vector9(0,0,0,0,-1,0,0,0,0), Vector9(0,0,0,Rij.s10,Rij.s11,Rij.s12,0,0,0)),
        Tuple2(Vector9(0,0,0,0,0,-1,0,0,0), Vector9(0,0,0,Rij.s20,Rij.s21,Rij.s22,0,0,0)),
        Tuple2(Vector9(0,0,0,0,0,0,-1,0,0), Vector9(0,0,0,0,0,0,Rij.s00,Rij.s01,Rij.s02)),
        Tuple2(Vector9(0,0,0,0,0,0,0,-1,0), Vector9(0,0,0,0,0,0,Rij.s10,Rij.s11,Rij.s12)),
        Tuple2(Vector9(0,0,0,0,0,0,0,0,-1), Vector9(0,0,0,0,0,0,Rij.s20,Rij.s21,Rij.s22))
      ])
```

Moreover, it becomes almost impossible to write code when I try to implement SVD for 3x3 rotation matrices as I cannot write stuff like `A[0, 0]`, and I cannot dynamically index the matrix in for loops... 

I am not sure if this design is good in a "Swifty" sense, so please correct me if I make silly mistakes!

marcrasi:
These operations sound good to me!

[...]

PR #107, opened 2020-06-26
quick and simple fixed size matrix

marcrasi:
Implements a `FixedSizeMatrix` supporting the operations described in https://github.com/borglab/SwiftFusion/issues/105.

This is pretty quick and dirty, just to get something useful working.

Let me know if it's missing anything or if there are more operations you'd like. Do you need differentiability? I'm happy to add anything you need :)

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -6,7 +6,7 @@
         "repositoryURL": "https://github.com/saeta/penguin.git",
         "state": {
           "branch": "master",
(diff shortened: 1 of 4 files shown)


ProfFan:
@marcrasi Thanks a lot :) I think for now differentiability is not needed, but it would definitely be a big plus, as I think in the foreseeable future we will be onto the land of differentiable algorithms (hopefully!)

PR #110, opened 2020-06-29
trap NaNs

marcrasi:
On Friday, we discussed the problem of figuring out where `NaN`s are happening. This is one of the solutions we discussed.

Adds a `TrappingDouble` that traps when there is a `NaN`, and typealiases `Double` to `TrappingDouble`. (Maybe this typealias should be commented out or controlled by an `#if` flag that is off by default, so that we don't have a weird nonstandard `Double` all the time. People debugging `NaN`s can enable the typealias locally.)

A few changes to other pieces of code were necessary to make this compile because `TrappingDouble` isn't a perfect drop in replacement for `Double`.

### DIFF
--- a/Sources/SwiftFusion/Core/TrappingDouble.swift
+++ b/Sources/SwiftFusion/Core/TrappingDouble.swift
@@ -25,12 +25,14 @@ public struct TrappingDouble: AdditiveArithmetic, Differentiable, Equatable {
   }
 }
 
+#if TRAPPING_DOUBLE
 /// A wrapper for `Double` that traps instead of allowing `NaN`.
 ///
 /// This typealias makes it easier to use unmodified existing code with `TrappingDouble`.
 ///
 /// If you really need the normal Swift `Double`, use `Swift.Double` to circumvent this typealias.
 public typealias Double = TrappingDouble
+#endif
 
 // The rest of this file provides conformances and functions that make the interface to
 // `TrappingDouble` very close to the interface to `Double`.
--- a/Sources/SwiftFusion/Core/TrappingDouble.swift
+++ b/Sources/SwiftFusion/Core/TrappingDouble.swift
@@ -1,0 +1,232 @@
+// Copyright 2020 The SwiftFusion Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import PythonKit
+import TensorFlow
+
+/// A wrapper for `Double` that traps instead of allowing `NaN`.
+public struct TrappingDouble: AdditiveArithmetic, Differentiable, Equatable {
+  public var value: Swift.Double
+
+  public init(_ value: Swift.Double) {
+    assert(!value.isNaN)
+    self.value = value
+  }
+}
+
+/// A wrapper for `Double` that traps instead of allowing `NaN`.
+///
+/// This typealias makes it easier to use unmodified existing code with `TrappingDouble`.
+///
+/// If you really need the normal Swift `Double`, use `Swift.Double` to circumvent this typealias.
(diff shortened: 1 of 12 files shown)


Sources/SwiftFusion/Core/TrappingDouble.swift
```diff
@@ -0,0 +1,232 @@
...
+    self.value = value
+  }
+}
+
+/// A wrapper for `Double` that traps instead of allowing `NaN`.
+///
+/// This typealias makes it easier to use unmodified existing code with `TrappingDouble`.
+///
+/// If you really need the normal Swift `Double`, use `Swift.Double` to circumvent this typealias.
+public typealias Double = TrappingDouble
```
ProfFan, line 33: I think we should have a RelWithAssertion build before globally enabling this?

ProfFan:
Wow! Awesome! I think now we can start debugging the NaNs (I think there is one when optimizing `sphere_bignoise`)

ProfFan:
@marcrasi I just tried using this to catch the NaN:

and in an instant, boom! here we are:
image

ProfFan:
@dellaert Since this is the same implementation as in GTSAM, maybe we have the same issue in GTSAM too?

Sources/SwiftFusion/Core/TrappingDouble.swift
marcrasi: Not sure what you mean. If we merge this now, then it'll trap NaNs on debug builds but not on release builds, which seems fine.

I do think we should comment out or `#if` out this line before merging, so that we don't have a weird nonstandard Double all the time. People debugging NaNs can enable the typealias locally.

ProfFan: Yup, my bad for the phrasing :) Let's `#if` out this before merging.

marcrasi:
> @marcrasi I just tried using this to catch the NaN:
> 
> and in an instant, boom! here we are:

That's cool! Do you know if this is happening in AD-transformed code? If it is, then you should see some functions named `AD__...` in the stack trace. These functions have source location information pointing at the original code, so it'll show the location of the original code even if it's one of the AD-transformed functions.

If this is happening in AD-generated code, it would also be interesting to know if it's happening in the forwards or backwards pass. I think you should be able to tell by looking if the AD functions are `AD_..._vjp_...` (forwards pass)` or `AD_..._pullback_...` (backwards pass).

ProfFan:
@marcrasi It's our code (both in a good and bad sense lol):

```swift
@differentiable(wrt: global)
  public func localCoordinate(_ global: Matrix3Coordinate) -> Vector3 {
    let relative = self.inverse() * global
    let R = relative.R
    let (R11, R12, R13) = (R.s00, R.s01, R.s02)
    let (R21, R22, R23) = (R.s10, R.s11, R.s12)
    let (R31, R32, R33) = (R.s20, R.s21, R.s22)

    let tr = R11 + R22 + R33

    if abs(tr + 1.0) < 1e-10 {
      ...
    } else {
      let tr_3 = tr - 3.0; // always negative
      if tr_3 < -1e-7 {
        // this is what needed to avoid the NaN
        // if tr <= -1.0 {
        //   let theta = Double.pi
        //   let magnitude = theta / (2.0 * sin(theta))
        //   return magnitude * Vector3(R32 - R23, R13 - R31, R21 - R12)
        // }
        let theta = acos((tr - 1.0) / 2.0) // Here is the NaN, tr is very slightly smaller than -1.0 because of error accumulation
        let magnitude = theta / (2.0 * sin(theta)) // in the optimization ( determinant becomes slightly offset from 1 )
        return magnitude * Vector3(R32 - R23, R13 - R31, R21 - R12)
      } else {
        // when theta near 0, +-2pi, +-4pi, etc. (trace near 3.0)
        // use Taylor expansion: theta \approx 1/2-(t-3)/12 + O((t-3)^2)
        let magnitude = (0.5 - tr_3 * tr_3 / 12.0)
        return magnitude * Vector3(R32 - R23, R13 - R31, R21 - R12)
      }
    }
```

ProfFan:
re-regularizing the rotation matrix to have `det=1` in every `move` could be expensive, so I think this "patch-up" fix is good for now.

ProfFan:
After this fix the NaNs are gone and I haven't observed any other NaNs, so great!

ProfFan:
Oops, need to update the download link to latest version for CI

marcrasi:
I think that the storage bucket with the toolchains has disappeared. I'm looking into it.

marcrasi:
We're looking into the vanished files. In the meantime, I've changed it to the v0.10 release toolchain which is pretty new and which hasn't vanished :)

PR #111, opened 2020-06-30
Fix the NaN in acos() of Rot3 localCoordinate

ProfFan:
This fixes the NaN error detected by #110 .

Kudos to @marcrasi for the awesome NaN detector :)


ProfFan:
Blame is on me :) turns out that I have one extra `abs()` in the Rot3 local code, so it goes into a wrong branch. Fixed that and now we are good to go.

PR #122, opened 2020-07-15
Replace Matrix3 with generic FixedSizeMatrix

marcrasi:
Replaces the hardcoded "Matrix3" with a generic "FixedSizeMatrix" that implements all the required operations generically.

The benchmark times are unaffected by this change, so the Swift compiler is specializing "FixedSizeMatrix" to code that is just as efficient as the hardcoded "Matrix3".

Next up in a subsequent PR: I'll import some LAPACK bindings so that we can use LAPACK operations on "FixedSizeMatrix", which should be much faster than doing the same operations on TensorFlow tensors. (Replacing the Rot3 SVD with this should speed up chordal initialization.)


Sources/SwiftFusion/Core/FixedSizeMatrix.swift
```diff
@@ -43,65 +47,312 @@ public struct FixedSizeMatrix<Rows: Equatable & FixedSizeArray> where Rows.Eleme
...
+}
+
+/// Returns the matrix-matrix product of `lhs` and `rhs`.
+///
+/// Note: This is currently only implemented for the product of square matrices of identical shape,
+/// but could be extended later.
+///
+/// - Requires: `type(of: lhs).isSquare`.
+@differentiable
+public func matmul<Rows>(_ lhs: FixedSizeMatrix<Rows>, _ rhs: FixedSizeMatrix<Rows>)
```
ProfFan, line 315: A naive impl of matmul would be just repeat matvec for n times, which would eliminate this code estate here? Feel free to ignore :)

Sources/SwiftFusion/Inference/BetweenFactorAlternative.swift
```diff
@@ -30,9 +30,7 @@ public struct BetweenFactorAlternative: LinearizableFactor2 {
     let R = actualMotion.coordinate.rot.coordinate.R + (-1) * difference.rot.coordinate.R
     let t = actualMotion.t - difference.t
     
-    // TODO(fan): Discuss with Marc why this would lead to failing CI
```
ProfFan, line 33: So is there an explanation for this? Just curious if it's me missing something or a compiler issue...

marcrasi: It looks like a compiler issue, maybe a bug in AutoDiff's handling of subscripts or switch statements. It's good that you caught it! I have filed a bug so that we can fix the underlying issue eventually: https://bugs.swift.org/browse/TF-1300.

Sources/SwiftFusion/Core/FixedSizeMatrix.swift
marcrasi: It's not easy to do it this way yet, because we don't have a function for forming the result matrix from column vectors. And the type of that function is nontrivial, so it might be a bit hard to add. But this definitely seems like a good thing to do after the method exists. (I imagine that we'll figure out how to add it eventually because it'll be very convenient for lots of things).

marcrasi:
https://github.com/borglab/SwiftFusion/pull/126 has revealed that this is a huge performance regression for the chordal initialization and chordal factor graph optimization (2x slower and 50x slower respectively).

I'll work on fixing the performance regression but @ProfFan if the performance regression is interfering with anything you're doing let me know and I could revert this PR until I have a fix for the regression.

ProfFan:
Thanks for the heads-up :) It’s not an issue at the moment

PR #128, opened 2020-07-17
Fix benchmarking instructions

dabrahams:


dabrahams:
@marcrasi I've taken to adding `--warmup-iterations 1 --iterations 10` so I can get some read on whether the results are stable.

PR #141, opened 2020-07-24
make it possible to use `Tensor` in factor graphs

marcrasi:
* Adds differentiable `init(flatTensor:)` and `.flatTensor` for converting fixed size vectors from/to `Tensor`, so that we can "escape" to `TensorFlow`-land in calculations that need `TensorFlow` functions.
* Adds `FixedShapeTensor` so that we can use `Tensor`s as variables and error vectors in factor graphs.

### DIFF
--- a/Sources/SwiftFusion/Core/FixedShapeTensor.swift
+++ b/Sources/SwiftFusion/Core/FixedShapeTensor.swift
@@ -1,0 +1,77 @@
+import TensorFlow
+
+public protocol FixedShapeTensor: EuclideanVectorN {
+  static var shape: TensorShape { get }
+  @differentiable init(_ tensor: Tensor<Double>)
+  @differentiable var tensor: Tensor<Double> { get set }
+}
+
+/// Default implementations of `EuclideanVectorN` requirements.
+extension FixedShapeTensor {
+  @differentiable
+  public static func += (_ lhs: inout Self, _ rhs: Self) {
+    lhs = Self(lhs.tensor + rhs.tensor)
+  }
+
+  @differentiable
+  public static func -= (_ lhs: inout Self, _ rhs: Self) {
+    lhs = Self(lhs.tensor - rhs.tensor)
+  }
+
+  @differentiable
+  public static func *= (_ lhs: inout Self, _ rhs: Double) {
+    lhs = Self(lhs.tensor * rhs)
+  }
+
+  @differentiable
+  public func dot(_ other: Self) -> Double {
+    (self.tensor * other.tensor).sum().scalarized()
+  }
+
+  public init<Source: Collection>(_ scalars: Source) where Source.Element == Double {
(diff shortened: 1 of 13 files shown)


dellaert:
@marcrasi, I'm so glad to have you aboard :-) This would have taken us ages to figure out!

marcrasi:
> @marcrasi, I'm so glad to have you aboard :-) This would have taken us ages to figure out!

I'm glad to help :)

marcrasi:
I've updated this with more docs and explanation so it should be ready for review & merge now.

Sources/SwiftFusion/Core/FixedShapeTensor.swift
```diff
@@ -0,0 +1,77 @@
+import TensorFlow
```
dellaert, line 1: Maybe provide some docs and rationale for why this module exists.

Sources/SwiftFusion/Core/FixedShapeTensor.swift
```diff
@@ -0,0 +1,77 @@
...
+    var scalars = self.tensor.scalars
+    let r = try scalars.withUnsafeMutableBufferPointer { b in
+      try body(b)
+    }
+    self.tensor = Tensor(shape: Self.shape, scalars: scalars)
+    return r
+  }
+}
+
+public struct Tensor10_10: AdditiveArithmetic, FixedShapeTensor {
```
dellaert, line 67: a bit weird to have just 10,10 here. Is that for a test?

ProfFan: It's for my PPCA, this is still very experimental, so I'll merge now and improve later.

marcrasi:
> where is flatTensor defined? Seems to suddenly pop up int his PR, but maybe I missed it

It's defined in this PR, in EuclideanVectorN.swift

Sources/SwiftFusion/Core/FixedShapeTensor.swift
marcrasi: currently `EuclideanVectorN` (which is used in all the factor graph infra) has a requirement that the size is known statically at compiletime. the easiest way to do this is to define separate types for all the shapes that we want to use. I have some simplifications in mind that might let us lift the restriction that the size be known at compiletime, so I'll try to remove this weird type by implementing those.

Sources/SwiftFusion/Core/FixedShapeTensor.swift
marcrasi: Ok, I'll do this in a follow up PR!

marcrasi:
Oopsie, I have just realized that I never actually pushed the commit with my doc comments before claiming that this was commented and ready for review! Sorry about that. I'll open a new PR with that commit.

PR #147, opened 2020-07-28
combine `EuclideanVector` & `EuclideanVectorN`, and rename them to `Vector`

### LINKED ISSUE #142: Rename `Vector` to `EuclideanVector` everywhere
The apparent (but non-existent) distinction is confusing.

marcrasi:
This is pretty much renaming with no functionality change. Now we have a single protocol `Vector` with all the requirements that are needed for vectors.

I needed to remove the dynamically-sized `Vector` and `Matrix` types to free up the names and also because the `Vector` protocol does not support dynamically-sized vectors.

In a subsequent PR, I want to change `Vector` to allow dynamically-sized vectors so that I can bring dynamically-sized vectors back. I want to make this change in a separate PR so that it is not overwhelmed by all the renaming changes.

Resolves #142.


Sources/SwiftFusion/Core/Vector.swift
```diff
@@ -1,162 +1,273 @@
+import Foundation
 import TensorFlow
 
-/// A dynamically sized vector.
-public struct Vector: Equatable, Differentiable {
-  public typealias TangentVector = Self
+/// An element of a Euclidean vector space.
```
dellaert, line 4: It might be good, now that your head is in this, to immediately split up this protocol and add a very simple test of a more general vector space. E.g., the space of polynomials that obeys it but is not a naked vector. Or, easier, add a comment what part of this protocol it would probably not conform to.

marcrasi: I have added comments about what parts it would probably not conform to.

I did not actually split the protocol because I'm not yet seeing where it would be useful. In fact, I have almost convinced myself that generalized vectors should conform to `Vector` even though `Vector` has additional structure. Reason: The actual representation of the vector in memory is going to be an array of scalars, and this representation "implies" a certain structure. Algorithms that technically don't depend on this structure (like matrix multiplication) nonetheless are most efficiently implemented by directly accessing elements of the array. Therefore algorithms should always have access to this array even when the vector is semantically a generalized vector.

Also, I have realized that `protocol Vector` has even more structure than Eucldiean. It also has a standard basis, used in the `.standardBasis` field, in the `init(_ scalars: Source)` method, and in the methods that let you access individual coordinates. I will update the comments to include this fact.

dabrahams:
Seems like if a fixed size is encoded into the protocol, now I can't make AnyVectorArrayBuffer conform to Vector, which is probably gonna make everything I'm doing more difficult.

marcrasi:
> Seems like if a fixed size is encoded into the protocol, now I can't make AnyVectorArrayBuffer conform to Vector, which is probably gonna make everything I'm doing more difficult.

I am working on removing the fixed size requirement. I wasn't able to remove it quite as quickly as I wanted, because a lot of things accidentally depend on the fixed size requirement, but I think all those things can be made to work without it.

PR #148, opened 2020-07-29
use dynamically allocated array for big jacobians

marcrasi:
My main motivation for this is so that we can have factors with big error vectors (on the order of 1,000 to 10,000 dimensions) without having to define fixed size arrays up to Array10000. We want to have error vectors with one element per pixel in an image.

Oddly, this also gave a 3% speedup in one of the benchmarks by changing the fixed size `Array6` to the dynamic `Array`. (See comment for details). Maybe it's avoiding some slow copies by passing around a pointer to the data.

This depends on a Penguin PR that I will create soon.

### DIFF
--- a/Sources/SwiftFusion/Inference/FactorsStorage.swift
+++ b/Sources/SwiftFusion/Inference/FactorsStorage.swift
@@ -200,9 +200,24 @@ extension AnyArrayBuffer where Dispatch == VectorFactorArrayDispatch {
     let dispatch: VectorFactorArrayDispatch
     let elementType = Type<Element>()
     typealias TangentVector = Element.LinearizableComponent.Variables.TangentVector
-    typealias Linearization<A: FixedSizeArray> = Type<JacobianFactor<A, Element.ErrorVector>>
-      where A.Element: Vector & DifferentiableVariableTuple
-    
+    typealias Linearization<A> = Type<JacobianFactor<A, Element.ErrorVector>>
+      where A: SourceInitializableCollection & UnsafeBufferAccessibleCollection,
+            A.Element: Vector & DifferentiableVariableTuple
+
+    // For small dimensions, we use a fixed size array in the linearization because the allocation
+    // and indirection overheads of a dynamically sized array are releatively big.
+    //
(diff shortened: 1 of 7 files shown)


Sources/SwiftFusion/Inference/FactorsStorage.swift
```diff
@@ -200,9 +200,24 @@ extension AnyArrayBuffer where Dispatch == VectorFactorArrayDispatch {
...
+    // new type and case for each size) and in some cases also happen to be faster than fixed size
+    // arrays.
+    //
+    // We chose 4 as the cutoff based on benchmark results:
+    // - "Pose2SLAM.FactorGraph", which heavily uses 3 dimensional error vectors, is ~30% faster
+    //   with a fixed size array than with a dynamically sized array.
+    // - "Pose3SLAM.FactorGraph", which heavily uses 6 dimensional error vectors, is ~3% faster
+    //   with a dynamically sized array than with a fixed size array.
+    // - "Pose3SLAM.sphere2500", which heavily uses 12 dimensional error vectors, has the same
+    //   performance with fixed size and dynamically sized arrays.
```
ProfFan, line 220: I think this is because we are not using SIMD and friends? I think if the loops are vectorized the fixed sized arrays should be a lot faster.

ProfFan: Probably should look at the generated IR and assembly to make sure.

ProfFan:
@marcrasi  Can we merge this and #153, #149 ? I think we can arrange a time for pair programming on these PRs and after finishing the PRs we can start again to finalize the PPCA :)

marcrasi:
> @marcrasi Can we merge this and #153, #149 ?

Yeah, sorry I've been working on all of them in parallel and not actually finishing any of them. If you want, I could just merge them all into `master` to unblock PPCA progress and then upload my work as subsequent PRs.

> I think we can arrange a time for pair programming on these PRs and after finishing the PRs we can start again to finalize the PPCA :)

I'm up for pair programming anytime today (but not in the next 30 minutes) or tomorrow after 2pm eastern time!

PR #150, opened 2020-07-30
Progress toward Differentiable AnyArrayBuffer

dabrahams:
Depends on #153.  

ArrayBuffers are differentiable and ArrayBuffers of Vector conform to Vector.

Does not yet make AnyArrayBuffer Differentiable, but all tests pass and the code is better, so I think we should accept in its current form.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -54,7 +54,10 @@ let package = Package(
       path: "Examples/Pose3SLAMG2O"),
     .testTarget(
       name: "SwiftFusionTests",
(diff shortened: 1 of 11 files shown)


dabrahams:
@marcrasi Please see https://github.com/borglab/SwiftFusion/pull/150/commits/74d6335676913842ec59589f6bf677b21a6aeb40 and https://github.com/borglab/SwiftFusion/pull/150/commits/f753af08f6cd95a7b341a6adaca07b3677998681
both of which suppress the crash.

Issue #152, opened 2020-08-03
remove as many dependencies on "fixed size vectors" as possible


marcrasi:
I'm about to make a PR that removes most dependencies on "fixed size vectors", but some of the changes would be better in a separate small PR, so here's a tracking issue for removing those.

PR #153, opened 2020-08-04
remove the requirement that vectors have fixed size

marcrasi:
This splits `Vector` into `Vector`, `ScalarsInitializableVector`, and `FixedSizeVector`, and updates most code to use the least constraining protocol.

There are a few more places to remove fixed size constraints, tracked in https://github.com/borglab/SwiftFusion/issues/152.


dellaert:
Cool progress. I am not in love with the name `ScalarsInitializableVector`. Can you explain more about the intent of the PR so I understand better or we can brainstorm another name? We can also do that on Friday

marcrasi:
> I am not in love with the name `ScalarsInitializableVector`. Can you explain more about the intent of the PR so I understand better or we can brainstorm another name?

I think I also don't like `ScalarsInitializableVector`. It's an unfortunate protocol that I'm forced to add because of the way I do some things. We may be able to get rid of it. Here's why it's forced to exist:

1. The matrix-vector multiplication algorithm dots each row of the matrix with the vector to produce a sequence of scalars. Then it uses `init(_ scalars)` to produce the result from the sequence of scalars. So we need `init(_ scalars)` in a protocol.
2. `init(_ scalars)` cannot go in `protocol FixedSizeVector` because then we would not be able to do mat-vec on dynamically sized vectors, which would defeat the purpose of this PR.
3. Tuples of `Vector`s conform to `Vector`. We cannot define `init(_ scalars)` for a tuple of dynamically-sized vectors because `init(_ scalars)` doesn't know how many scalars to put in each component of the tuple. Therefore, `init(_ scalars)` cannot go in `protocol Vector`.
4. The "easy way" out of this dilemma is to add a new `protocol ScalarsInitializableVector` between `protocol Vector` and `protocol FixedSizeVector`.

One way to get rid of `ScalarsInitializableVector` would be to eliminate (3) by removing the conformance of tuples of `Vector`s to `Vector`. This change is nontrivial because `JacobianFactor` depends heavily on the conformance, so I should try to make the change in a separate PR. I think you would also like this change for another reason: currently `JacobianFactor` has a single matrix whose rows are tuples, which you previously expressed some concern about. The change would make it so that `JacobianFactor` would have a tuple of matrices, one per adjacent variable.

dabrahams:
Part of the problem with the name is just the use of the plural “scalars;” if it was “ScalarInitializableVector” that would make it more grammatical.  But the name also reveals that it's a weak concept.  Anything with a standard basis, that's also mutable, should be initializable with scalars.

dellaert:
> The change would make it so that JacobianFactor would have a tuple of matrices, one per adjacent variable.

I love that, indeed :-)

PR #158, opened 2020-08-07
Make MCMC tests stop failing nondeterministically.

dabrahams:
Use a pseudo-random number generator instead of the system one.


dabrahams:
Fixes #143

dellaert:
Thanks @dabrahams - sorry, I was pre-occupied with other things :-/

PR #161, opened 2020-08-11
faster crop

marcrasi:
This makes the crop faster in a few ways:
* Do the crop on an `Array`, which has faster indexing than a `Tensor`.
* Do the derivative only with respect to the bounding box position, so that we don't spend time calculating the derivative with respect to the input image.
* Do the pullback with a custom vjp that is more efficient than the synthesized vjp.

@ProfFan this is just the speedups that I have already shown you. I have some additional speedups that I'll put in a separate PR soon.


dellaert:
Nice !

PR #162, opened 2020-08-12
remove `ScalarsInitializableVector`

marcrasi:
Most uses of `ScalarsInitializableVector` only happen on concrete types, so we can move that functionality to concrete types.

One use in `JacobianFactor.swift` is truly generic but we can implement that by mutating the `zero` vector instead (like we did with standard basis).

### DIFF
--- a/Sources/SwiftFusion/Core/FixedShapeTensor.swift
+++ b/Sources/SwiftFusion/Core/FixedShapeTensor.swift
@@ -38,10 +38,6 @@ extension FixedShapeTensor {
     (self.tensor * other.tensor).sum().scalarized()
   }
 
-  public init<Source: Collection>(_ scalars: Source) where Source.Element == Double {
-    self.init(Tensor(shape: Self.shape, scalars: Array(scalars)))
-  }
-
   public var dimension: Int {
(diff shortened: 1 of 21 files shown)


dabrahams (changes requested):
Well see, if you can implement that initialization in O(N) by creating the zero vector and mutating it, it shows that this is not a separate protocol from `Vector`.  However, there's also no reason not to give `Vector` that `init` requirement with a default implementation that mutates the zero vector.  Then you can make a more efficient implementation available wherever it's convenient.

marcrasi:
> Well see, if you can implement that initialization in O(N) by creating the zero vector and mutating it, it shows that this is not a separate protocol from Vector. However, there's also no reason not to give Vector that init requirement with a default implementation that mutates the zero vector. Then you can make a more efficient implementation available wherever it's convenient.

Interesting. If we can implement it with the zero vector and mutation, then this does seem good, but I've now realized that it isn't always right to implement it that way, if we want to support dynamically-sized vectors.

I thought about vectors a lot yesterday and I'm overall confused about the right way forward so I'll just wait until we talk before trying to make any more changes.

PR #139, opened 2020-07-24
[WIP] PPCA

ProfFan:
Currently blocked by not having non-square matrix ops.

PR #163, opened 2020-08-13
faster linearization for PPCATrackingFactor

marcrasi:
Adds a `PPCATrackingFactor` like in https://github.com/borglab/SwiftFusion/pull/139, but with a faster linearization that computes the Jacobian using a hand coded forwards mode derivative.

The time it takes to linearize goes from ~2000ms to ~1ms. The time it takes to run CGLS given one of these linearizations stays constant around 2ms.

There are tests that the custom linearization gives the same value as the autodiff linearization, *but no tests that either linearization works correctly*, because those are the experiments that @ProfFan is working on.

This PR adds a quick and dirty "if" statement to the factor graph implementation to dispatch to the custom linearization, but this could eventually become a feature allowing anyone to write a custom linearization.

@ProfFan, feel free to schedule some time on my calendar if you want to do some pair coding to talk about this, to integrate this with your experiments, to work more on your experiments, or whatever :)


dellaert:
Wow! How many times can you do a PR with a 2000x runtime improvement :-) Will look at it now!

Sources/SwiftFusion/Inference/PPCATrackingFactor.swift
```diff
@@ -0,0 +1,119 @@
+import PenguinStructures
+import TensorFlow
+
+/// A factor over a target's pose and appearance in an image.
+public struct PPCATrackingFactor: LinearizableFactor2 {
+  /// The first adjacent variable, the pose of the target in the image.
+  public typealias V0 = Pose2
```
dellaert, line 7: V0 and V1 are not very intuitive names :-) Can we call them Pose and Latent?

Tests/SwiftFusionTests/Inference/PPCATrackingFactorTests.swift
```diff
@@ -0,0 +1,87 @@
...
+import PenguinStructures
+@testable import SwiftFusion
+
+class PPCATrackingFactorTests: XCTestCase {
+  /// Tests that the custom `linearized(at:)` method produces the same results at autodiff.
+  func testLinearizedValue() {
+    // Initialize `factor` with random fields.
+    let image = Tensor<Double>(randomNormal: [500, 500, 1])
+    let W = Tensor<Double>(
+      randomNormal: PPCATrackingFactor.Patch.shape + [PPCATrackingFactor.V1.dimension])
```
dellaert, line 28: I know this PR is about the speedup, but since I'm reading to code in detail (and it has been a while) allow me to spam it with general thoughts about the overall design. 

In particular: It seems odd to have to relate to a *factor* here. We might want to refactor (no pun intended!) so we have a *PPCA generative model* that is separate from the factor, which would use it. That sets up a good pattern for the VAE iteration.

Tests/SwiftFusionTests/Inference/PPCATrackingFactorTests.swift
```diff
@@ -0,0 +1,87 @@
...
+    let mu = PPCATrackingFactor.Patch(
+      Tensor<Double>(randomNormal: PPCATrackingFactor.Patch.shape))
```
dellaert, line 30: Also, a PPCA model would not necessarily be specific to images, but can model arbitrary tensors. So, having `Patch` here also feels odd.

Tests/SwiftFusionTests/Inference/PPCATrackingFactorTests.swift
```diff
@@ -0,0 +1,87 @@
...
+    let poseID = TypedID<Pose2>(0)
+    let latentID = TypedID<Vector5>(0)
+    let factor = PPCATrackingFactor(poseID, latentID, measurement: image, W: W, mu: mu)
```
dellaert, line 33: that's awesome :-)
Proposed refactor might pass in a PPCA model rather than W and mu separately.

Tests/SwiftFusionTests/Inference/PPCATrackingFactorTests.swift
```diff
@@ -0,0 +1,87 @@
...
+
+    for _ in 0..<2 {
+      let linearizationPoint = Tuple2(
+        Pose2(randomWithCovariance: eye(rowCount: 3)),
```
dellaert, line 37: Oh man! thats awesome!

Tests/SwiftFusionTests/Inference/PPCATrackingFactorTests.swift
```diff
@@ -0,0 +1,87 @@
...
+        Vector5(flatTensor: Tensor(randomNormal: [5])))
+
+      typealias Variables = PPCATrackingFactor.Variables.TangentVector
+      typealias ErrorVector = PPCATrackingFactor.ErrorVector
+
+      let autodiff = JacobianFactor<Array<Variables>, ErrorVector>(
```
dellaert, line 43: Might want to add a comment (in addition to method comment, which says this already)
```
// Below we compare custom and much faster differentiation with the autodiff version
```

Tests/SwiftFusionTests/Inference/PPCATrackingFactorTests.swift
```diff
@@ -0,0 +1,87 @@
...
+        linearizing: factor, at: linearizationPoint)
+      let custom = factor.linearized(at: linearizationPoint)
+
+      XCTAssertEqual(
+        custom.errorVector(at: Variables.zero), autodiff.errorVector(at: Variables.zero))
+
+      for _ in 0..<10 {
```
dellaert, line 50:
```
// Check the linear part of the derivative at 10 random points for the latent variable v
```

BTW, is this deterministic? @dabrahams will frown if not :-)

Tests/SwiftFusionTests/Inference/PPCATrackingFactorTests.swift
```diff
@@ -0,0 +1,87 @@
...
+        let v = Variables(flatTensor: Tensor(randomNormal: [Variables.dimension]))
```
dellaert, line 51: In a refactor, we might say
```
a = ppca.randomLatent()
```
Note I think `a` is a better name than v: it matches the paper, stands for appearance in the bee case.

Tests/SwiftFusionTests/Inference/PPCATrackingFactorTests.swift
```diff
@@ -0,0 +1,87 @@
...
+        assertEqual(
+          custom.errorVector_linearComponent(v).tensor,
+          autodiff.errorVector_linearComponent(v).tensor,
+          accuracy: 1e-6)
+      }
+
+      for _ in 0..<10 {
```
dellaert, line 58: Add a similar comment, I assuming this is the reverse AD pass that is being tested

Tests/SwiftFusionTests/Inference/PPCATrackingFactorTests.swift
```diff
@@ -0,0 +1,87 @@
...
+          accuracy: 1e-6)
+      }
+    }
+  }
+
+  /// Tests that factor graphs with a `PPCATrackingFactor`s linearize them using the custom
+  /// linearization.
+  func testFactorGraphUsesCustomLinearized() {
+    let image = Tensor<Double>(randomNormal: [500, 500, 1])
+    let W = Tensor<Double>(
```
dellaert, line 72: Duplicate, can we use a fixture?

Tests/SwiftFusionTests/Inference/PPCATrackingFactorTests.swift
```diff
@@ -0,0 +1,87 @@
...
+
+    var x = VariableAssignments()
+    let poseId = x.store(Pose2(100, 100, 0))
+    let latentId = x.store(Vector5.zero)
+
+    var fg = FactorGraph()
+    fg.store(PPCATrackingFactor(poseId, latentId, measurement: image, W: W, mu: mu))
+
+    let gfg = fg.linearized(at: x)
+    XCTAssertNotNil(gfg.storage.first!.value[elementType: Type<LinearizedPPCATrackingFactor>()])
```
dellaert, line 85: if you say so ! No idea what's going on here (yet). Comment warranted

Sources/SwiftFusion/Image/Patch.swift
```diff
@@ -69,14 +69,59 @@ extension ArrayImage {
...
+    var du = ArrayImage(rows: region.rows, cols: region.cols, channels: self.channels)
+    var dv = ArrayImage(rows: region.rows, cols: region.cols, channels: self.channels)
+    for i in 0..<region.rows {
+      for j in 0..<region.cols {
+        // The position of the destination pixel in the destination image, in `(u, v)` coordinates.
+        let uvDest = Vector2(Double(j) + 0.5, Double(i) + 0.5)
+
+        // The position of the destination pixel in the source image, in coordinates where the
+        // center of the destination image is `(0, 0)`.
+        let xyDest = uvDest - 0.5 * Vector2(Double(region.cols), Double(region.rows))
```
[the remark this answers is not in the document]
marcrasi: Oops, there is a mistake in the comment. This is still a position in the destination image (not the "source image" as said in the comment), so it doesn't need to be rotated yet. I'll fix the comment.

Sources/SwiftFusion/Image/Patch.swift
```diff
@@ -87,18 +132,46 @@ extension Tensor where Scalar == Double {
     let result = ArrayImage(self).patch(at: region).tensor
     return self.shape.count == 2 ? result.reshaped(to: [region.rows, region.cols]) : result
   }
+
+  /// Returns the patch of `self` at `region`, and its Jacobian with respect to the center of the
```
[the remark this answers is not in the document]
marcrasi: The `center` field of the region is a `Pose2` with the pose of its center. I see how it sounds like it's just a 2d position though. I'll change this comment and also think if there is a better name for the `center` field.

Sources/SwiftFusion/Image/Patch.swift
```diff
@@ -87,18 +132,46 @@ extension Tensor where Scalar == Double {
...
+  ///                the directional derivatives along the rotation and the `u`, `v` translation
+  ///                components (in that order).
+  public func patchWithJacobian(at region: OrientedBoundingBox)
+    -> (patch: Tensor<Scalar>, jacobian: Tensor<Scalar>)
+  {
+    precondition(self.shape.count == 3, "image must have shape [height, width, channelCount]")
+    let (patch, jacobian) = ArrayImage(self).patchWithJacobian(at: region)
+    return (
+      patch: patch.tensor,
+      jacobian: Tensor<Double>(stacking: [
```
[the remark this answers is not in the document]
marcrasi: I expect that the value semantics does not magically optimize this because the behavior of `Tensor(stacking:)` is probably very opaque to the compiler.

The reason I'm doing it in two steps is that `Tensor` does not support any efficient way of writing elements into it (raw memory access or otherwise). It is much faster to build an array and then copy it into a tensor than to write elements to the tensor directly.

Of course this is slower than writing directly so we should switch it to writing directly if/when we have a tensor that lets us do that efficiently.

Another possibility might be those precomputed Jacobians we talked about last meeting. It sounds like those would involve a small number of tensor operations that the tensor backend would (hopefully) handle efficiently. The precomputed Jacobians are out of scope of this PR though.

Sources/SwiftFusion/Inference/FactorsStorage.swift
```diff
@@ -214,31 +240,37 @@ extension AnyArrayBuffer where Dispatch == VectorFactorArrayDispatch {
...
+      // arrays.
+      //
+      // We chose 4 as the cutoff based on benchmark results:
+      // - "Pose2SLAM.FactorGraph", which heavily uses 3 dimensional error vectors, is ~30% faster
+      //   with a fixed size array than with a dynamically sized array.
+      // - "Pose3SLAM.FactorGraph", which heavily uses 6 dimensional error vectors, is ~3% faster
+      //   with a dynamically sized array than with a fixed size array.
+      // - "Pose3SLAM.sphere2500", which heavily uses 12 dimensional error vectors, has the same
+      //   performance with fixed size and dynamically sized arrays.
+      switch Element.ErrorVector.dimension {
```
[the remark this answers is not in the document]
marcrasi: This is existing code that I indented.

It's code that decides how to store the linearization of a factor based on the factor's error vector.

Sources/SwiftFusion/Inference/PPCATrackingFactor.swift
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  /// Returns a linear approximation to `self` at `x`.
+  public func linearized(at x: Variables) -> LinearizedPPCATrackingFactor {
+    let pose = x.head
+    let latent = x.tail.head
+
+    let bbox = OrientedBoundingBox(center: pose, rows: Patch.shape[0], cols: Patch.shape[1])
+    let generatedAppearance =
+      mu + Patch(matmul(W, latent.flatTensor.expandingShape(at: 1)).squeezingShape(at: 2))
```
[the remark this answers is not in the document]
marcrasi: When linearizing a factor graph, we only run the `linearized` method because it calculates everything required for the linearization, including the error vector. So it's not duplicate work.

This comment did make me notice that this is duplicate code, so I'll factor it out into another method.

Sources/SwiftFusion/Inference/PPCATrackingFactor.swift
marcrasi: These are explicitly specifying `LinearizableFactor2`'s associatedtype `V0` and `V1` so they have to be named like this. I have added a comment explaining that.

Sources/SwiftFusion/Inference/PPCATrackingFactor.swift
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  @differentiable
+  public func errorVector(at x: Variables) -> PPCATrackingFactor.Patch {
+    errorVector_linearComponent(x) - error
+  }
+
+  public func errorVector_linearComponent(_ x: Variables) -> PPCATrackingFactor.Patch {
+    PPCATrackingFactor.Patch(
+      matmul(jacobian, x.flatTensor.expandingShape(at: 1)).squeezingShape(at: 2))
```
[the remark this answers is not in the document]
marcrasi: I think this sounds good. I will start doing this. Ideally one day we'll never have to write chain rule jacobians because autodiff will do it for us, but since I'm writing some chain rule jacobians now I'll start adopting this convention :)

Sources/SwiftFusion/Inference/PPCATrackingFactor.swift
dabrahams: Well, but `V0` and `V1` are not awesome associated type names either, though I'm not sure we can do better.  I guess `V` stands for `Variable`?  I think if we named them `Variable0`, etc., or `AdjacentVariable0`, or `Neighbor0` the code might be more self-documenting.  The only question is, would other code become difficult for clients?

Tests/SwiftFusionTests/Inference/PPCATrackingFactorTests.swift
dabrahams: Could you say more about what seems odd to you here, Frank?

marcrasi: I see that templating it could be good.

I think we should start out doing copy and paste though because there are some pieces that are not completely trivial to abstract over (like custom jacobian) and it'll be easier to abstract over them once we have a few examples.

Tests/SwiftFusionTests/Inference/PPCATrackingFactorTests.swift
dabrahams: I think this is an indicator that a higher level API on gfg is needed, encapsulating this test.  Though the test looks awfully specific.  GFG shouldn't be exposing that dictionary in the long run.

Frank, this is testing that the first (key,value) pair in the dictionary underlying gfg has a value that is an `AnyArrayBuffer` containing `LinearizedPPCATrackingFactor` elements.

Sources/SwiftFusion/Inference/PPCATrackingFactor.swift
marcrasi: I think that `Variable0` etc sounds really good. No client code uses these anywhere as far as I know, so it shouldn't cause any changes except making the declaration easier to read.

(In theory you don't need to declare these at all because associated type inference figures them out automatically. But in practice the error messages when associated type inference fails are so incomprehensible that I think we should really encourage every factor implementation to explicitly write them out.)

marcrasi: I'll update the names in a separate PR though.

Tests/SwiftFusionTests/Inference/PPCATrackingFactorTests.swift
marcrasi: I shall add seeds.

marcrasi:
Thanks for the comments! I've addressed/responded to them. I'll merge soon.

PR #164, opened 2020-08-14
Add Vector.scalars

dabrahams:
This does not yet remove accesses to vectors through withUnsafeMutableBufferPointer, but now that scalars are required to be supported, we are in a position to remove that from the protocol. 

No change in performance.

### DIFF
--- a/Sources/SwiftFusion/Inference/ArrayBuffer+Vector.swift
+++ b/Sources/SwiftFusion/Inference/ArrayBuffer+Vector.swift
@@ -13,7 +13,97 @@
 // limitations under the License.
 import PenguinStructures
 
+public struct NestedIndex<OuterIndex: Comparable, InnerIndex: Comparable> : Comparable {
+  fileprivate var outer: OuterIndex
+  fileprivate var inner: InnerIndex?
+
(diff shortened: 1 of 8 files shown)


Sources/SwiftFusion/Inference/ArrayBuffer+Vector.swift
```diff
@@ -13,7 +13,97 @@
...
+    else {
+      self.init(endIn: c)
+    }
+  }
+  
+  internal init<C: Collection>(firstValidIn c: C)
+    where C.Index == OuterIndex, C.Element: Collection, C.Element.Index == InnerIndex
+  {
+    self.init(firstValidIn: c) { $0 }
+  }
```
marcrasi, line 50: This one is unused.

dabrahams, line 50:
```suggestion
```

dabrahams: I know; it all needs comments and tests, but this is a convenience for the case where you have a direct collection-of-collections, with no need for projection to get the inner collection.

dabrahams:
Superseded by #165

PR #165, opened 2020-08-14
Differentiable scalars

dabrahams:
All the comments are in and this is ready to land now.

### DIFF
--- a/Sources/SwiftFusion/Core/Vector.swift
+++ b/Sources/SwiftFusion/Core/Vector.swift
@@ -52,7 +52,7 @@ public protocol Vector: Differentiable where Self.TangentVector == Self {
   // the standard basis.
 
   #if true // set to false to find examples to be removed.
-  // DWA TODO: Remove these requirements!
+  // TODO(marcrasi): Remove these requirements!
   
   /// Returns the result of calling `body` on the scalars of `self`.
   ///
@@ -141,7 +141,7 @@ extension Vector {
 }
 
 #if true // set to false to find examples to be removed.
-// DWA TODO: Remove these requirements!
+// TODO(marcrasi): Remove these implementations!
 
 /// Default implementations of raw memory access.
 extension Vector {
--- a/Sources/SwiftFusion/Core/Vector.swift
+++ b/Sources/SwiftFusion/Core/Vector.swift
@@ -8,6 +8,10 @@ import TensorFlow
 // have Euclidean structure and standard basis, so we'll reserve the short name `Vector` for
 // such vectors and use a longer name like `GeneralizedVector` for the generalized vector spaces.
 public protocol Vector: Differentiable where Self.TangentVector == Self {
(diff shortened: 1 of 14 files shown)


Sources/SwiftFusion/Core/Vector.swift
```diff
@@ -1,15 +1,21 @@
 import Foundation
 import TensorFlow
 
+protocol DifferentiableMutableCollection: Differentiable, MutableCollection {
+  @differentiable
+  var count: Int { get }
```
marcrasi, line 6: This should not be `@differentiable` because it's not a `Differentiable` type.

If the compiler is complaining about `x.count` not being differentiable, you can do `withoutDerivative(at: x).count`.

marcrasi: However, I think that all the methods using `.count` (`.flatTensor`) are ones whose implementations are not automatically differentiable and therefore the real solution is to make sure that the custom derivative implementation is registered correctly.

dan-zheng: > If the compiler is complaining about `x.count` not being differentiable, you can do `withoutDerivative(at: x).count`.

Alternatively, you can mark `var count` as `@noDerivative`. That achieves the same affect as `withoutDerivative(at: x).count` without complicating use sites, you can continue to use `x.count`.

Sources/SwiftFusion/Core/Vector.swift
```diff
@@ -45,6 +51,9 @@ public protocol Vector: Differentiable where Self.TangentVector == Self {
   // MARK: - Methods for accessing the standard basis and for manipulating the coordinates under
   // the standard basis.
 
+  #if true // set to false to find examples to be removed.
+  // DWA TODO: Remove these requirements!
```
marcrasi, line 55: Change to `TODO(marcrasi)`?

Sources/SwiftFusion/Core/Vector.swift
```diff
@@ -130,6 +140,9 @@ extension Vector {
   }
 }
 
+#if true // set to false to find examples to be removed.
+// DWA TODO: Remove these requirements!
```
marcrasi, line 144: Change to `TODO(marcrasi)`?

Sources/SwiftFusion/Core/Vector.swift
dabrahams: When you have change suggestions, please use the “±” button in the web interface to propose them.  It makes them ever so much easier to accept!

dabrahams, line 55:
```suggestion
  // TODO(marcrasi): Remove these requirements!
```

Sources/SwiftFusion/Core/Vector.swift
dabrahams, line 144:
```suggestion
// TODO(marcrasi): Remove these implementations!
```

PR #172, opened 2020-08-18
AnyArrayBuffer : Differentiable

dabrahams:


Sources/SwiftFusion/Inference/AnyArrayBuffer+Differentiable.swift
```diff
@@ -0,0 +1,85 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/// Contiguous storage for factor graph values (variable assignments, error vectors) of statically
+/// unknown type.
+
```
marcrasi, line 17: This file comment looks wrong because this isn't the file that declares the storage types. Maybe just delete it?

```suggestion
```

PR #175, opened 2020-08-20
Eliminate as many uses of `static var zero` as possible

dabrahams:
3.3% speedup on Pose3SLAM benchmark.
1.1% on Pose2SLAM

### DIFF
--- a/Sources/SwiftFusion/Core/Dictionary+Differentiable.swift
+++ b/Sources/SwiftFusion/Core/Dictionary+Differentiable.swift
@@ -16,7 +16,7 @@ extension Dictionary: Differentiable where Value: Differentiable {
   }
 
   public var zeroTangentVectorInitializer: () -> TangentVector {
-    { mapValues { v in v.zeroTangentVectorInitializer() } }
+    { mapValues { v in v.zeroTangentVector } }
   }
 }
 
--- a/Sources/SwiftFusion/Core/Dictionary+Differentiable.swift
+++ b/Sources/SwiftFusion/Core/Dictionary+Differentiable.swift
@@ -16,7 +16,7 @@ extension Dictionary: Differentiable where Value: Differentiable {
   }
 
   public var zeroTangentVectorInitializer: () -> TangentVector {
-    { mapValues { _ in .zero } }
+    { mapValues { v in v.zeroTangentVectorInitializer() } }
   }
 }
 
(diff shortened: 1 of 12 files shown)


Sources/SwiftFusion/Core/Dictionary+Differentiable.swift
```diff
@@ -16,7 +16,7 @@ extension Dictionary: Differentiable where Value: Differentiable {
   }
 
   public var zeroTangentVectorInitializer: () -> TangentVector {
-    { mapValues { _ in .zero } }
+    { mapValues { v in v.zeroTangentVectorInitializer() } }
```
marcrasi, line 19: `Differentiable` has a default computed field `.zeroTangentVector` that returns `.zeroTangentVectorInitializer()`.

Sources/SwiftFusion/Core/FixedSizeMatrix.swift
```diff
@@ -325,7 +322,7 @@ extension FixedSizeMatrix {
 @differentiable
 public func matvec<Rows>(_ lhs: FixedSizeMatrix<Rows>, _ rhs: Rows.Element) -> Rows.Element {
   precondition(type(of: lhs).isSquare, "matvec only implemented for square matrices")
-  var r = Rows.Element.zero
+  var r = Rows.Element(repeatElement(0.0, count: rhs.dimension))
```
marcrasi, line 332: Using `repeatElement(0.0, count: rhs.dimension)` won't work to initialize zero vectors of all types (like the dictionary-like heterogeneous vector arrays).

Those will need something like `.init(zerosCompatibleWith:)`.

But this PR is already an improvement so we could make that improvement later in a separate PR.

(This is similar to my comment in https://github.com/borglab/SwiftFusion/pull/176#discussion_r474169145).

Sources/SwiftFusion/Core/Dictionary+Differentiable.swift
dabrahams: And you suggest I use that instead?

marcrasi: Yes, I think it's easier to understand. But if you disagree then I'm fine keeping it as `.zeroTangentVectorInitializer()`.

dabrahams: OK, I just pushed a change that replaces all of those.  Arguably this is a 3.2% speedup (go figure)

Sources/SwiftFusion/Core/FixedSizeMatrix.swift
dabrahams: Yes, one step at a time please.

dabrahams: Also, how is such a vector going to be the row of a `FixedSizeMatrix`?

PR #167, opened 2020-08-17
test vector semantics

marcrasi:
Adds a `checkVectorSemantics` method that checks the recently-added `Scalars` and `scalars` requirements.

Migrates all the existing vector semantics checks to be inside of `checkVectorSemantics` too.

PR #176, opened 2020-08-20
Factor out standardBasis as a lazy collection

dabrahams:


Sources/SwiftFusion/Core/Vector.swift
```diff
@@ -92,6 +92,37 @@ public protocol ScalarsInitializableVector: Vector {
...
+  /// Scalar values of the created instance are all zero except for the `n`th one, which is 1.
+  ///
+  /// - Requires: `n < m`
+  public init(standardBasisMember n: Int, of m: Int) {
+    precondition(n < m)
+    self.init(repeatElement(0.0, count: m))
+    // As of this writing, mutating is faster than initializing with the right value
+    scalars[scalars.index(atOffset: n)] = 1.0
+  }
+}
```
marcrasi, line 107: This initializer won't work on vector types that cannot be initialized given a just dimension. (Like the heterogeneous vector arrays).

A `standardBasis` instance property would work in that case though (like I just did in https://github.com/borglab/SwiftFusion/pull/167):

```swift
  /// The standard basis vectors with the same shape as `self`.
  public var standardBasis: LazyMapCollection<Scalars.Indices, Self> {
    self.scalars.indices.lazy.map { i in
      var r = self.zeroTangentVector
      r.scalars[i] = 1
      return r
    }
  }
```

This also produces the whole standard basis faster in cases where `scalars.index(atOffset: n)` is `O(n)`.

What do you think about doing it this way instead?

dabrahams: OK, I pushed a commit that changes the way this is done.  Right now `standardBasis` remains a `static var` on `FixedSizeVector`, but the parts could be used for any `Vector` given an instance.

dabrahams: Probably needs tests…

PR #180, opened 2020-09-03
Remove more nondeterminism in MCMC test

dabrahams:
Addresses, but by itself does not fix, #143.  Currently, the test fails deterministically.

@dellaert please take a look and make sure the test is correct.

### DIFF
--- a/Tests/SwiftFusionTests/MCMC/MCMCTests.swift
+++ b/Tests/SwiftFusionTests/MCMC/MCMCTests.swift
@@ -15,7 +15,7 @@ class MCMCTests: XCTestCase {
   /// Sampling from the Standard Normal Distribution.
   /// Inspired by testRWM1DUniform from tfp
   func testRWM1DUniform() throws {
-    var deterministicEntropy = ARC4RandomNumberGenerator(seed: 42)
+    let deterministicEntropy = ARC4RandomNumberGenerator(seed: 42)
     
     // Create kernel for MCMC sampler:
     // target_log_prob_fn is proportional to log p(x), where p is a zero-mean Gaussian
@@ -23,8 +23,8 @@ class MCMCTests: XCTestCase {
     // in this case perturbing x with uniformly sampled perturbations.
     let kernel = RandomWalkMetropolis(
       sourceOfEntropy: deterministicEntropy,
-      target_log_prob_fn: {(x:Double) in -0.5*x*x}, //  tfd.Normal(loc=dtype(0), scale=dtype(1))
-      new_state_fn: {(x:Double) in x + Double.random(in: -1..<1, using: &deterministicEntropy)}
+      target_log_prob_fn: {(x: Double) in -0.5*x*x}, //  tfd.Normal(loc=dtype(0), scale=dtype(1))
+      new_state_fn: {(x: Double, r: inout AnyRandomNumberGenerator) in x + Double.random(in: -1..<1, using: &r)}
     )
     
     // Run the sampler for 2500 steps, discarding the first 500 asa burn-in
--- a/Tests/SwiftFusionTests/MCMC/MCMCTests.swift
+++ b/Tests/SwiftFusionTests/MCMC/MCMCTests.swift
@@ -15,14 +15,16 @@ class MCMCTests: XCTestCase {
   /// Sampling from the Standard Normal Distribution.
   /// Inspired by testRWM1DUniform from tfp
   func testRWM1DUniform() throws {
+    var deterministicEntropy = ARC4RandomNumberGenerator(seed: 42)
+    
     // Create kernel for MCMC sampler:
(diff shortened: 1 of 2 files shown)


ProfFan:
Bump @dellaert

dellaert:
> Bump @dellaert

Master no longer compiles on my Mac (Xcode 11.7 on Catalina). What's the intel on how to make this work?

ProfFan:
@dellaert Probably you need the latest XCode beta: https://github.com/borglab/SwiftFusion/pull/181#issuecomment-693108106

dellaert:
So @dabrahams, the test seems correct. And I investigated the difference between the built-in random generator and the ARC4 one, and I can't see a meaningful difference of random-sampling from [-1,1]:

image

**Yet** the test fails *for every seed I tried* with ARC4, with a value for around -1 for the mean rather than zero. I'm stumped for now: brainstorm what could be wrong?

ProfFan:
I think I know why. There is a copy operation in the constructor, so the algorithm is bifurcating a random number generator without changing the seed. I can try push a fix...

dellaert:
Yeah, just add a commit to the PR. I have no outstanding changes in the branch.

ProfFan:
I think this is a problem about the container type `AnyRNG`. It is a type erased container, but it creates a copy, and it does not allow dereferencing the inner object...

Tests/SwiftFusionTests/MCMC/MCMCTests.swift
```diff
@@ -15,16 +15,16 @@ class MCMCTests: XCTestCase {
...
     // Create kernel for MCMC sampler:
     // target_log_prob_fn is proportional to log p(x), where p is a zero-mean Gaussian
     // new_state_fn is a symmetric (required for Metropolis) proposal density,
     // in this case perturbing x with uniformly sampled perturbations.
     let kernel = RandomWalkMetropolis(
       sourceOfEntropy: deterministicEntropy,
-      target_log_prob_fn: {(x:Double) in -0.5*x*x}, //  tfd.Normal(loc=dtype(0), scale=dtype(1))
-      new_state_fn: {(x:Double) in x + Double.random(in: -1..<1, using: &deterministicEntropy)}
+      target_log_prob_fn: {(x: Double) in -0.5*x*x}, //  tfd.Normal(loc=dtype(0), scale=dtype(1))
+      new_state_fn: {(x: Double, r: inout AnyRandomNumberGenerator) in x + Double.random(in: -1..<1, using: &r)}
```
ProfFan, line 27: @dabrahams Do you think this is a good pattern?

dellaert: Well, that won't work without changing the Metropolis sampler. 

But, I am not convinced this is the problem: I obtained the histograms above by adding a print statement inside `new_state_fn`, and printing out the 2500 random numbers generated. So, it's definitely not the case that I got the same number every time...

ProfFan: You get the same number *twice*, IMHO, so the distribution looks exactly the same

dellaert: > You get the same number _twice_, IMHO, so the distribution looks exactly the same

I don't understand

dellaert (approved):
Fan and I looked into it and I think with @ProfFan 's latest commit this is now fixed :-) Thanks @ProfFan !!!

Tests/SwiftFusionTests/MCMC/MCMCTests.swift
dabrahams: @ProfFan I'm not sure it's a great pattern, because [here](https://github.com/borglab/SwiftFusion/blob/master/Sources/SwiftFusion/MCMC/RandomWalkMetropolis.swift#L34) the instance of `RandomWalkMetropolis` is storing its source of entropy (after wastefully wrapping it in another `AnyRandomNumberGenerator`).  Therefore, the two instances of `RandomWalkMetropolis` will see the same sequence of random numbers, but shifted.  I guess that's the case with all PRNGs, but in this case the shift will be by the same number of samples each time the test is run.  Do you think that's alright?  It might be the case that we should create a class wrapper to give `ARC4RandomNumberGenerator` reference semantics.

dabrahams: I take it all back (except for the wasteful wrapping on the linked line, which should be eliminated) my brain was addled by doing infrastructure work in Python; that's my excuse anyway.

dellaert: What should the line be instead? I can create a PR

dabrahams: Ugh, never mind; I was wrong about that, too! 😭 
dave.credibility = 0️⃣

PR #181, opened 2020-09-03
Make AnyArrayBuffer conform to Vector

dabrahams:
There's plenty I'd like to improve in here, plenty to discuss, and plenty of tests to add, but I wanted to get this in front of @marcrasi  ASAP

### DIFF
--- a/Sources/SwiftFusion/Inference/FlattenedScalars.swift
+++ b/Sources/SwiftFusion/Inference/FlattenedScalars.swift
@@ -1,0 +1,154 @@
+// TODO: Move all these to Penguin.
+
+import PenguinStructures
+
+public struct FlattenedScalars<Base: MutableCollection>
+  where Base.Element: Vector
+{
+  var base: Base
+    
+  init(_ base: Base) {
+    self.base = base
+  }
+}
+
+extension FlattenedScalars: Sequence {
+  public typealias Element = Base.Element.Scalars.Element
+  
+  public struct Iterator: IteratorProtocol {
+    var outer: Base.Iterator
+    var inner: Base.Element.Scalars.Iterator?
+
+    init(outer: Base.Iterator, inner: Base.Element.Scalars.Iterator? = nil) {
+      self.outer = outer
+      self.inner = inner
+    }
+
+    public mutating func next() -> Element? {
+      while true {
+        if let r = inner?.next() { return r }
+        guard let o = outer.next() else { return nil }
+        inner = o.scalars.makeIterator()
+      }
+    }
+  }
+
+  public func makeIterator() -> Iterator { Iterator(outer: base.makeIterator()) }
+
+  // https://bugs.swift.org/browse/SR-13486 points out that the withContiguousStorageIfAvailable
+  // method is dangerously underspecified, so we don't implement it here.
+  
+  public func _customContainsEquatableElement(_ e: Element) -> Bool? {
+    let m = base.lazy.map({ $0.scalars._customContainsEquatableElement(e) })
+    if let x = m.first(where: { $0 != false }) {
+      return x
+    }
+    return false
+  }
+
+  public __consuming func _copyContents(
+    initializing t: UnsafeMutableBufferPointer<Element>
+  ) -> (Iterator, UnsafeMutableBufferPointer<Element>.Index) {
+    var r = 0
+    var outer = base.makeIterator()
+
+    while r < t.count, let e = outer.next() {
+      var (inner, r1) = e.scalars._copyContents(
+        initializing: .init(start: t.baseAddress.map { $0 + r }, count: t.count - r))
+      r += r1
+
+      // See if we ran out of target space before reaching the end of the inner collection.  I think
+      // this will be rare, so instead of using an O(N) `e.count` and comparing with `r1` in all
+      // passes of the outer loop, spend `inner` seeing if we reached the end and reconstitute it in
+      // O(N) if not.
+      if inner.next() != nil {
+        @inline(never)
+        func earlyExit() -> (Iterator, UnsafeMutableBufferPointer<Element>.Index) {
+          var i = e.scalars.makeIterator()
+          for _ in 0..<r1 { _ = i.next() }
+          return (.init(outer: outer, inner: i), r)
+        }
+        return earlyExit()
+      }
+    }
+    return (.init(outer: outer, inner: nil), r)
+  }
+}
+
+extension FlattenedScalars: MutableCollection {
+  public typealias Index = FlattenedIndex<Base.Index, Base.Element.Scalars.Index>
+
+  public var startIndex: Index { .init(firstValidIn: base, innerCollection: \.scalars) }
+  
+  public var endIndex: Index { .init(endIn: base) }
+  
+  public func index(after x: Index) -> Index {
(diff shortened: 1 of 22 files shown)


marcrasi:
I shall look later today! I'm working on a few urgent issues first, but I plan to get to this next.

dabrahams:
@marcrasi This works for me with `TOOLCHAINS=com.google.swift.20200826`.  Is there any way we can upgrade the bots that are testing SwiftFusion?

marcrasi:
> @marcrasi This works for me with TOOLCHAINS=com.google.swift.20200826. Is there any way we can upgrade the bots that are testing SwiftFusion?

It's configured in https://github.com/borglab/SwiftFusion/blob/master/.github/workflows/ci.yml

dabrahams:
Any ideas?  I'm stumped.

ProfFan:
@marcrasi  We should move to a Linux-based CI I think, macOS env is not consistent on Actions...

marcrasi:
Yeah, maybe the macos version is too old for the new Swift toolchain.

marcrasi:
Sorry, the urgent stuff took up all my time today. I'll try again to look at this tomorrow :)

Sources/SwiftFusion/Inference/FlattenedScalars.swift
```diff
@@ -0,0 +1,154 @@
...
+      }
+    }
+  }
+
+  public func makeIterator() -> Iterator { Iterator(outer: base.makeIterator()) }
+
+  // https://bugs.swift.org/browse/SR-13486 points out that the withContiguousStorageIfAvailable
+  // method is dangerously underspecified, so we don't implement it here.
+  
+  public func _customContainsEquatableElement(_ e: Element) -> Bool? {
```
marcrasi, line 41: Wow, there are lots of secret stdlib protocol requirements. I don't know what they are for but I'd be curious to hear more about them.

Sources/SwiftFusion/Inference/PenguinExtensions.swift
```diff
@@ -104,6 +137,577 @@ extension ArrayBuffer: CustomDebugStringConvertible {
...
+public struct FlattenedMutableCollection<Base: MutableCollection>
+  where Base.Element: MutableCollection
+{
+  var base: Base
+    
+  init(_ base: Base) {
+    self.base = base
+  }
+}
+
```
marcrasi, line 556: Could this somehow be used to implement `FlattenedScalars`?

If this used a closure `(Base.Element) -> InnerCollection` to extract the inner collection then `FlattenedScalars` could be implemented as a `FlattenedMutableCollection` with `\.scalars` for the closure. Though I guess that probably wouldn't optimize well.

dabrahams: It certainly *might* not optimize well, and it's not a chance I wanted to take.

Sources/SwiftFusion/Inference/PenguinExtensions.swift
```diff
@@ -104,6 +137,577 @@ extension ArrayBuffer: CustomDebugStringConvertible {
...
+    vtable[0].copyContents(self, target)
+  }
+}
+
+extension AnyMutableCollection: MutableCollection {
+  public struct Index: Comparable {
+    fileprivate var storage: Storage
+    // Note: using a vtable here actually made benchmarks slower.  We might want to try again if we
+    // expand what the benchmarks are doing.  My hunch is that the cost of copying an Index is less
+    // important than the cost of checking for equality.
```
marcrasi, line 439: It's not immediately obvious to me why the vtable makes checking for equality slower. Could you explain why? Is it: The vtable makes the equality checking slower because it has to follow an extra pointer every time it makes the function call?

Sources/SwiftFusion/Inference/PenguinExtensions.swift
```diff
@@ -104,6 +137,577 @@ extension ArrayBuffer: CustomDebugStringConvertible {
...
+    initializing target: UnsafeMutableBufferPointer<Element>
+  ) -> (Iterator,UnsafeMutableBufferPointer<Element>.Index) {
+    vtable[0].copyContents(self, target)
+  }
+}
+
+extension AnyMutableCollection: MutableCollection {
+  public struct Index: Comparable {
+    fileprivate var storage: Storage
+    // Note: using a vtable here actually made benchmarks slower.  We might want to try again if we
```
marcrasi, line 437: There aren't any benchmarks in this PR. Do you have some benchmarks of this that you have been running?

marcrasi (commented):
I finished reading through this.

Sources/SwiftFusion/Inference/FlattenedScalars.swift
dabrahams: Happy to talk them through with you whenever

Sources/SwiftFusion/Inference/PenguinExtensions.swift
```diff
@@ -104,6 +137,577 @@ extension ArrayBuffer: CustomDebugStringConvertible {
...
+  ) rethrows {
+    let r = try self_[unsafelyAssuming: Type<C>()]
+      .partition { try withUnsafePointer(to: $0) { try belongsInSecondPartition($0) } }
+    resultIndex.assumingMemoryBound(to: AnyMutableCollection<C.Element>.Index?.self)
+      .pointee = .init(r)
+  }
+}
+
+/// A type-erased mutable collection similar to the standard library's `AnyCollection`, but with
+/// parameterized type-erasure substrate.
```
dabrahams, line 267: Honestly, I have no idea what I meant by that!
```suggestion
/// A type-erased mutable collection similar to the standard library's `AnyCollection`.
```

Sources/SwiftFusion/Inference/PenguinExtensions.swift
dabrahams: \*sigh\*, you people and your benchmarks! 😉 

The benchmarks for this thing are essentially covered here:
https://github.com/saeta/penguin/blob/master/Sources/Benchmarks/AnyValueBenchmarks.swift#L300-L302
and when this thing becomes public Penguin API they should be expanded.  I don't think benchmarks for this component, which we're going to move to penguin, belong in SwiftFusion.

As for the other parts, I'm not sure I know how to benchmark them.  I'd been assuming that when we get them wired into SwiftFusion we'll observe their performance through existing benchmarks, but if you have other ideas, I'm all ears.

Sources/SwiftFusion/Inference/PenguinExtensions.swift
dabrahams: Yes exactly.  If you have an idea of how to clarify that in the comment, please leave a suggestion.

dabrahams:
@marcrasi I think the bot needs the latest Xcode-beta installed (and xcode-select called to point at it).  Do you think you could set that up?

marcrasi:
I searched around for github actions documentation and found this: https://github.com/actions/virtual-environments/blob/main/images/macos/macos-10.15-Readme.md

Looks like we might be able to get xcode beta only a `xcode-select`. I'll try that and if it doesn't work I'll see if I can do something else that works.

marcrasi:
It worked!!! 🎉

marcrasi (approved):

PR #182, opened 2020-09-09
Improve group actions

tuwuhs:
`Rot2`:

- [x] Make public
- [x] Create group actions in `Rot2Coordinate`
- [x] Move group actions to a separate extension

`Rot3`:

- [x] Make public
- [x] Move group actions to a separate extension

`Pose2`:

- [x] Move group action in `Pose2Coordinate` to a separate extension

`Pose3`:

- [x] Create group action

### DIFF
--- a/Sources/SwiftFusion/Geometry/Pose2.swift
+++ b/Sources/SwiftFusion/Geometry/Pose2.swift
@@ -85,8 +85,8 @@ extension Pose2 {
 extension Pose2 {
   /// Group action on `Vector2`.
   @differentiable
-  public static func * (lhs: Pose2, rhs: Vector2) -> Vector2 {
-    lhs.coordinate * rhs
+  public static func * (aTb: Pose2, bp: Vector2) -> Vector2 {
+    aTb.coordinate * bp
   }
 }
 
@@ -138,8 +138,8 @@ extension Pose2Coordinate: LieGroupCoordinate {
 extension Pose2Coordinate {
   /// Group action on `Vector2`.
   @differentiable
-  static func * (lhs: Pose2Coordinate, rhs: Vector2) -> Vector2 {
-    lhs.t + lhs.rot * rhs
+  static func * (aTb: Pose2Coordinate, bp: Vector2) -> Vector2 {
+    aTb.rot * bp + aTb.t
   }
 }
 
--- a/Sources/SwiftFusion/Geometry/Pose2.swift
+++ b/Sources/SwiftFusion/Geometry/Pose2.swift
@@ -138,7 +138,7 @@ extension Pose2Coordinate: LieGroupCoordinate {
 extension Pose2Coordinate {
   /// Group action on `Vector2`.
   @differentiable
-  public static func * (lhs: Pose2Coordinate, rhs: Vector2) -> Vector2 {
+  static func * (lhs: Pose2Coordinate, rhs: Vector2) -> Vector2 {
     lhs.t + lhs.rot * rhs
   }
 }
--- a/Sources/SwiftFusion/Geometry/Pose2.swift
+++ b/Sources/SwiftFusion/Geometry/Pose2.swift
@@ -133,7 +133,9 @@ extension Pose2Coordinate: LieGroupCoordinate {
   public func inverse() -> Pose2Coordinate {
     Pose2Coordinate(self.rot.inverse(), self.rot.unrotate(-self.t))
   }
+}
 
+extension Pose2Coordinate {
   /// Group action on `Vector2`.
   @differentiable
   public static func * (lhs: Pose2Coordinate, rhs: Vector2) -> Vector2 {
--- a/Sources/SwiftFusion/Geometry/Pose3.swift
+++ b/Sources/SwiftFusion/Geometry/Pose3.swift
@@ -35,6 +35,14 @@ public struct Pose3: LieGroup, Equatable, KeyPathIterable {
   }
 }
 
+extension Pose3 {
+  /// Group action on `Vector3`.
+  @differentiable
+  public static func * (_ lhs: Pose3, _ rhs: Vector3) -> Vector3 {
+    lhs.coordinate * rhs
+  }
+}
+
 // MARK: - Global Coordinate System
 
 public struct Pose3Coordinate: Equatable, KeyPathIterable {
@@ -94,6 +102,14 @@ extension Pose3Coordinate: LieGroupCoordinate {
   @differentiable
   public func inverse() -> Pose3Coordinate {
     Pose3Coordinate(self.rot.inverse(), self.rot.unrotate(-self.t))
+  }
+}
+
+extension Pose3Coordinate {
+  /// Group action on `Vector3`.
+  @differentiable
+  static func * (_ lhs: Pose3Coordinate, _ rhs: Vector3) -> Vector3 {
+    lhs.t + lhs.rot * rhs
   }
 }
 
--- a/Sources/SwiftFusion/Geometry/Pose3.swift
+++ b/Sources/SwiftFusion/Geometry/Pose3.swift
@@ -38,8 +38,8 @@ public struct Pose3: LieGroup, Equatable, KeyPathIterable {
 extension Pose3 {
   /// Group action on `Vector3`.
   @differentiable
-  public static func * (_ lhs: Pose3, _ rhs: Vector3) -> Vector3 {
-    lhs.coordinate * rhs
+  public static func * (_ aTb: Pose3, _ bp: Vector3) -> Vector3 {
+    aTb.coordinate * bp
   }
 }
 
@@ -108,8 +108,8 @@ extension Pose3Coordinate: LieGroupCoordinate {
 extension Pose3Coordinate {
   /// Group action on `Vector3`.
   @differentiable
-  static func * (_ lhs: Pose3Coordinate, _ rhs: Vector3) -> Vector3 {
-    lhs.t + lhs.rot * rhs
+  static func * (_ aTb: Pose3Coordinate, _ bp: Vector3) -> Vector3 {
+    aTb.rot * bp + aTb.t
   }
 }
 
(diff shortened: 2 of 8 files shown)


Sources/SwiftFusion/Geometry/Pose2.swift
```diff
@@ -133,10 +133,12 @@ extension Pose2Coordinate: LieGroupCoordinate {
   public func inverse() -> Pose2Coordinate {
     Pose2Coordinate(self.rot.inverse(), self.rot.unrotate(-self.t))
   }
+}
 
+extension Pose2Coordinate {
   /// Group action on `Vector2`.
   @differentiable
-  public static func * (lhs: Pose2Coordinate, rhs: Vector2) -> Vector2 {
+  static func * (lhs: Pose2Coordinate, rhs: Vector2) -> Vector2 {
```
marcrasi, line 141: Did you remove `public` intentionally? I think this should be public.

Sources/SwiftFusion/Geometry/Rot2.swift
```diff
@@ -48,23 +48,25 @@ extension Rot2: CustomDebugStringConvertible {
   }
 }
 
+/// Group actions
```
marcrasi, line 51: The standard style is to always end doc comments with a period, even when it's not a complete sentence (https://swift.org/documentation/api-design-guidelines/ > Write a documentation comment "More details" > Begin with a summary "More details")

Sources/SwiftFusion/Geometry/Rot3.swift
```diff
@@ -51,23 +51,26 @@ public struct Rot3: LieGroup, Equatable, KeyPathIterable {
       txy+twz, 1.0-(txx+tzz), tyz-twx,
       txz-twy, tyz+twx, 1.0-(txx+tyy)
     )
-  }
-  
+  }  
+}
+
+/// Group actions
```
marcrasi, line 57: Same comment about a period.

marcrasi (changes requested):
Thanks! Looks great to me with just a few nits

Sources/SwiftFusion/Geometry/Pose2.swift
marcrasi: Oh actually I see that it is the "Coordinate". Non-public seems good.

tuwuhs: Yeah, I had been thinking whether or not to declare it public, since from what I gathered the "...Coordinate" is actually an implementation detail of the actual type.

Sources/SwiftFusion/Geometry/Rot2.swift
tuwuhs: Done.

Sources/SwiftFusion/Geometry/Rot3.swift
tuwuhs: Done.

Sources/SwiftFusion/Geometry/Pose3.swift
```diff
@@ -97,6 +105,14 @@ extension Pose3Coordinate: LieGroupCoordinate {
   }
 }
 
+extension Pose3Coordinate {
+  /// Group action on `Vector3`.
+  @differentiable
+  static func * (_ lhs: Pose3Coordinate, _ rhs: Vector3) -> Vector3 {
```
dellaert, line 111: So, I get why you would use `lhs` and `rhs`, but I would propose changing all these methods to use the "coordinate system memnonic", as extra documentation and sanity check. In GTSAM we write a pose as 'a_pose_b' where a refers to the target frame A and b to the source frame B. We then write `a_point = a_pose_b * b_point`, and you can see that indices cancel, as they would in tensor contraction @dabrahams :-)

Same for rotations, so below it would read as:
```
a_rotation_b * b_point + a_translation
```

Note also I *deliberately* choose Rotation/translation order, and *deliberately* don't abbreviate. I think of a 3D pose as [R,t], not [t,R]. Let's make our code think the same way :-)

Sources/SwiftFusion/Geometry/Rot2.swift
```diff
@@ -48,23 +48,25 @@ extension Rot2: CustomDebugStringConvertible {
...
 }
 
+/// Group actions.
 extension Rot2 {
   /// Returns the result of acting `self` on `v`.
   @differentiable
-  func rotate(_ v: Vector2) -> Vector2 {
-    Vector2(c * v.x - s * v.y, s * v.x + c * v.y)
+  public func rotate(_ v: Vector2) -> Vector2 {
+    coordinate.rotate(v)
```
dellaert, line 56: Just a comment: @marcrasi I still think it's weird we have this coordinate business :-/ It's like we double up everything? I know I can probably be convinced again, but just conveying my gut feeling here that it might look weird to outsiders. @tuwuhs what was your impression of this? We're genuinely interested in your first impression here.

Tests/SwiftFusionTests/Geometry/Pose2Tests.swift
```diff
@@ -356,4 +356,37 @@ final class Pose2Tests: XCTestCase {
...
+
+    assertAllKeyPathEqual(actual, expected, accuracy: 1e-9)
+  }
+
+  /// Tests group action: rotate only
+  func testActionRotate() {
+    let x = Vector2(1.0, 2.0)
+    let pose = Pose2(0.0, 0.0, 0.5)
+
+    let expected = Vector2(-0.08126851531803325,  2.2345906623849485)
```
dellaert, line 376: Make tests easy and obvious :-) Use an angle like 30 or 60 degrees for which cos and sine are well known. Write sqrt() if you need to. don't use 45 as its special.

Tests/SwiftFusionTests/Geometry/Pose2Tests.swift
```diff
@@ -356,4 +356,37 @@ final class Pose2Tests: XCTestCase {
...
+
+    assertAllKeyPathEqual(actual, expected, accuracy: 1e-9)
+  }
+
+  /// Tests group action: translate and rotate
+  func testActionTranslateRotate() {
+    let x = Vector2(1.0, -2.0)
+    let pose = Pose2(-3.0, 4.0, 0.5)
+
+    let expected = Vector2(-1.1635663609012212, 2.7242604148234575)
```
dellaert, line 387: same

Tests/SwiftFusionTests/Geometry/Pose3Tests.swift
```diff
@@ -156,4 +156,37 @@ final class Pose3Tests: XCTestCase {
...
+
+    assertAllKeyPathEqual(actual, expected, accuracy: 1e-9)
+  }
+
+  /// Tests group action: rotate only
+  func testActionRotate() {
+    let x = Vector3(1.0, -2.0, 3.0)
+    let pose = Pose3(Rot3.fromTangent(Vector3(0.1, 0.2, -0.3)), Vector3(0, 0, 0))
+
+    let expected = Vector3(0.871509606555838, -2.5663299211301474, 2.5796165880985145)
```
dellaert, line 176: same, although it might be hard to find a "beautiful" rotation in 3D. I'd have one test with a single yaw pitch or roll which establishes trust, and then a "regression test" with these hardcoded numbers. How do I know these hardcoded numbers are correct in the first place? Comment that :-)

dellaert (changes requested):
Cool Tuwuh ! Welcome to the team! 

And apologies in advanced for my anal retentive reviewing style :-)

Sources/SwiftFusion/Geometry/Rot2.swift
tuwuhs: I think this coordinate thing does make sense for rotations, since we can have more than one way to implement a rotation. For instance `Rot3` can be implemented either as a rotation matrix or a quaternion. For rigid transforms I agree that it seems like a lot of code duplications with no clear advantage (at least in my limited experience working with Swift and Lie groups). What is the main rationale behind separating the coordinate type? Is it some subtle Swifty thing? Or perhaps something to do with differentiability?

Tests/SwiftFusionTests/Geometry/Pose2Tests.swift
tuwuhs: Got it, will do.

Sources/SwiftFusion/Geometry/Pose3.swift
tuwuhs: Yes, `b_point` makes it clearer that the point is currently in the source frame B and we want to move it to the target frame A. I don't actually have a hard opinion on this though. For me the semantics is already quite clear and unambiguous as it is.

So, what about something like this?
```swift
static func * (_ a_pose_b: Pose3Coordinate, _ b_point: Vector3) -> Vector3 {
    a_pose_b.rot * b_point + a_pose_b.t
}
```

Or should we go with a more verbose version?
```swift
static func * (_ a_pose_b: Pose3Coordinate, _ b_point: Vector3) -> Vector3 {
    let a_rotation_b = a_pose_b.rot
    let a_translation = a_pose_b.t
    return a_rotation_b * b_point + a_translation
}
```

Also, these break the naming convention rule, is it okay?

Btw, what is tensor contraction? Any primer on the topic to help me get started?

dellaert: If naming convention is camelCase, then let's use aTb, aRb etc. I do prefer the less verbose version above.

Tests/SwiftFusionTests/Geometry/Pose3Tests.swift
tuwuhs: Done.

Tests/SwiftFusionTests/Geometry/Pose2Tests.swift
tuwuhs: Done.

Tests/SwiftFusionTests/Geometry/Pose2Tests.swift
tuwuhs: Done.

Sources/SwiftFusion/Geometry/Pose3.swift
tuwuhs: Done.

dellaert (approved):
Excellent!

marcrasi (approved):

PR #186, opened 2020-09-18
Eliminate a wasteful wrapping layer

dabrahams:


dellaert:
So, this fails to compile for me (and seemingly for the checks as well).
Xcode suggested
```
self.sourceOfEntropy = sourceOfEntropy as! AnyRandomNumberGenerator
```
but that segfaults.

dabrahams:
Because this suggestion is totally bogus.  What can I say, I was having a bad day.  Sorry!

PR #187, opened 2020-09-22
fix Package.swift

marcrasi:
A dependency published a new version that prevented swiftpm dependency resolution from working. I patched swift-models to fix the dependency, and this PR updates SwiftFusion to used the patched swift-models.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -6,7 +6,7 @@
         "repositoryURL": "https://github.com/saeta/penguin.git",
         "state": {
           "branch": "master",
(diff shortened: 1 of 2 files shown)


marcrasi:
Actually this doesn't fix the problem, never mind. There is a different way to deconfuse SwiftPM.

If you do get the problem:
```
error: the Package.resolved file is most likely severely out-of-date and is preventing correct resolution; delete the resolved file and try again
```

Then you can fix it by running `swift build`. (The error message's suggestion of deleting Package.resolved does not work for me.)

PR #192, opened 2020-10-10
add a gradient descent factor graph optimizer

marcrasi:
I included a test on a simple factor graph, but I haven't tried to use it on anything more complicated like the PPCA tracking factor graph. I would expect it to work without any problems.


ProfFan (approved):
Thanks Marc :)

PR #193, opened 2020-10-15
Pinhole camera first draft

tuwuhs:
Here's my first draft of the pinhole camera implementation. I based the structure on GTSAM. Some notes:
- Personally I feel like the name `Cal3_S2` is a bit confusing. I think the `3` and `2` seem redundant if we consider a camera to be a mapping from 3D world to 2D image, unless we extend the definition of a camera. Also it isn't immediately apparent what the `_S` part means. Maybe change it to something like `SimpleCalibration` or `CalSimple` or `CalSimpleWithSkew`?
- Not all calibration model has a closed-form inverse, especially for the distortion part. Most models are defined in the `uncalibrate()` direction from normalized coordinate to image coordinate. Requiring `calibrate()` in `CameraCalibration` means that some implementations may need to find solution through iteration. It feels like an additional layer of complexity hidden from user. However, `calibrate()` is needed to implement `backproject()`. I just think it's worth mentioning so that we are aware of the potential issue.

Let me know what you guys think!

### DIFF
--- a/Sources/SwiftFusion/Geometry/Cal3_S2.swift
+++ b/Sources/SwiftFusion/Geometry/Cal3_S2.swift
@@ -34,15 +34,15 @@ public struct Cal3_S2: CameraCalibration, Equatable {
 /// CameraCalibration protocol conformance.
 extension Cal3_S2 {
   @differentiable
-  public func uncalibrate(_ pNormalized: Vector2) -> Vector2 {
+  public func uncalibrate(_ np: Vector2) -> Vector2 {
     Vector2(
-      fx * pNormalized.x + s * pNormalized.y + u0,
-      fy * pNormalized.y + v0)
+      fx * np.x + s * np.y + u0,
+      fy * np.y + v0)
   }
 
   @differentiable
-  public func calibrate(_ pImage: Vector2) -> Vector2 {
-    let (du, dv) = (pImage.x - u0, pImage.y - v0)
+  public func calibrate(_ ip: Vector2) -> Vector2 {
+    let (du, dv) = (ip.x - u0, ip.y - v0)
     let (fxInv, fyInv) = (1.0 / fx, 1.0 / fy)
     return Vector2(
       fxInv * du - s * fxInv * fyInv * dv,
--- a/Sources/SwiftFusion/Geometry/Cal3_S2.swift
+++ b/Sources/SwiftFusion/Geometry/Cal3_S2.swift
@@ -1,6 +1,6 @@
 
 /// The five-parameter camera calibration.
-public struct Cal3_S2: CameraCalibration, Manifold, Equatable {
+public struct Cal3_S2: Manifold, Equatable {
   public typealias Coordinate = Cal3_S2Coordinate
   public typealias TangentVector = Vector5
 
@@ -25,8 +25,8 @@ public struct Cal3_S2: CameraCalibration, Manifold, Equatable {
   }
 }
 
-/// CameraCalibration protocol conformance.
-extension Cal3_S2 {
+/// CameraCalibration conformance.
+extension Cal3_S2: CameraCalibration {
   @differentiable
   public func uncalibrate(_ np: Vector2) -> Vector2 {
     coordinate.uncalibrate(np)
@@ -38,9 +38,9 @@ extension Cal3_S2 {
   }
 }
 
-public struct Cal3_S2Coordinate: ManifoldCoordinate, Equatable {
-  public typealias LocalCoordinate = Vector5
 
+/// Manifold coordinate for Cal3_S2.
+public struct Cal3_S2Coordinate: Equatable {
   /// Focal length in X direction.
   public var fx: Double
 
@@ -85,6 +85,21 @@ public struct Cal3_S2Coordinate: ManifoldCoordinate, Equatable {
   }
 }
 
+/// ManifoldCoordinate conformance.
+extension Cal3_S2Coordinate: ManifoldCoordinate {
+  public typealias LocalCoordinate = Vector5
+
+  @differentiable(wrt: local)
+  public func retract(_ local: Vector5) -> Cal3_S2Coordinate {
+    Cal3_S2Coordinate(asVector() + local)
+  }
+
+  @differentiable(wrt: global)
+  public func localCoordinate(_ global: Cal3_S2Coordinate) -> Vector5 {
+    global.asVector() - asVector()
+  }
+}
+
 /// Operations on a point.
 extension Cal3_S2Coordinate {
   @differentiable
@@ -101,18 +116,5 @@ extension Cal3_S2Coordinate {
     return Vector2(
       fxInv * du - s * fxInv * fyInv * dv,
       fyInv * dv)
-  }
-}
-
-/// ManifoldCoordinate conformance.
-extension Cal3_S2Coordinate {
-  @differentiable(wrt: local)
-  public func retract(_ local: Vector5) -> Cal3_S2Coordinate {
-    Cal3_S2Coordinate(asVector() + local)
-  }
-
-  @differentiable(wrt: global)
-  public func localCoordinate(_ global: Cal3_S2Coordinate) -> Vector5 {
-    global.asVector() - asVector()
   }
 }
--- a/Sources/SwiftFusion/Geometry/Cal3_S2.swift
+++ b/Sources/SwiftFusion/Geometry/Cal3_S2.swift
@@ -104,17 +104,13 @@ extension Cal3_S2Coordinate: ManifoldCoordinate {
 extension Cal3_S2Coordinate {
   @differentiable
   public func uncalibrate(_ np: Vector2) -> Vector2 {
-    Vector2(
-      fx * np.x + s * np.y + u0,
-      fy * np.y + v0)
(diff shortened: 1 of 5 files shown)


ProfFan:
I don't have bandwidth for this right now, but I think `Cal3_S2` should conform to `LieGroup` just like `Pose3`.

tuwuhs:
I think conforming to `Manifold` would make more sense, since we don't compose two calibrations.
In fact making the `Cal3_S2` conform to `ManifoldCoordinate` might be even better, since:
- The tangent vector will be a specific type instead of just a `Vector5`, improving type safety
- We can still make a `Manifold` out of it for doing optimization

Something along the lines of
```swift
public struct Cal3_S2: CameraCalibration, Equatable, ManifoldCoordinate {
  // ...

  public func asVector() -> Vector5 {
    Vector5(fx, fy, s, u0, v0)
  }

  @differentiable(wrt: local)
  public func retract(_ local: Vector5) -> Cal3_S2 {
    CameraCalibration(asVector() + local)
  }

  @differentiable(wrt: global)
  public func localCoordinate(_ global: Cal3_S2) -> Vector5 {
    global.asVector() - asVector()
  }
}

public struct Cal3_S2Manifold: Equatable, Differentiable, Manifold {
  public typealias Coordinate = Cal3_S2
  public typealias TangentVector = Vector5
  public var coordinateStorage: Cal3_S2

  // ...

  public mutating func move(along direction: Vector5) {
    coordinateStorage = coordinateStorage.retract(direction)
  }
}
```

ProfFan:
Note that `ManifoldCorrdinate` is a workaround for some S4TF limitations (for now). Please refer to what `Pose3` is doing with `Pose3Coordinate`. In short, `Foo` conforming to `Manifold` should have a inner `FooCoordinate`.

tuwuhs:
Oh, I didn't know that. OK then I'll add the `Manifold` conformance.

Sources/SwiftFusion/Geometry/Cal3_S2.swift
```diff
@@ -0,0 +1,51 @@
+
+/// The five-parameter camera calibration.
+public struct Cal3_S2: CameraCalibration, Equatable {
+  /// Focal length in X direction.
```
dellaert, line 44: not: make this more compact by doing one-line comments after definition, and getting rid of newlines

Sources/SwiftFusion/Geometry/Cal3_S2.swift
```diff
@@ -0,0 +1,51 @@
...
+  public init() {
+    self.init(fx: 1.0, fy: 1.0, s: 0.0, u0: 0.0, v0: 0.0)
+  }
+}
+
+/// CameraCalibration protocol conformance.
+extension Cal3_S2 {
+  @differentiable
+  public func uncalibrate(_ np: Vector2) -> Vector2 {
+    Vector2(
```
dellaert, line 107: one-liner?

Sources/SwiftFusion/Geometry/Cal3_S2.swift
```diff
@@ -0,0 +1,51 @@
...
+      fx * np.x + s * np.y + u0,
+      fy * np.y + v0)
+  }
+
+  @differentiable
+  public func calibrate(_ ip: Vector2) -> Vector2 {
+    let (du, dv) = (ip.x - u0, ip.y - v0)
+    let (fxInv, fyInv) = (1.0 / fx, 1.0 / fy)
+    return Vector2(
```
dellaert, line 116: one-liner? Do you use a Swift-formatter? I think we conform to Apple style

Sources/SwiftFusion/Geometry/CameraCalibration.swift
```diff
@@ -0,0 +1,14 @@
+
+/// A type for camera calibration parameters.
```
dellaert, line 2: a protocol

Sources/SwiftFusion/Geometry/PinholeCamera.swift
```diff
@@ -0,0 +1,104 @@
...
+
+        return (
+          TangentVector(wTc: dpose, calibration: calibration.zeroTangentVector),
+          dpoint)
+      }
+    )
+  }
+
+  /// Projects a 3D point in the world frame to 2D normalized coordinate and returns intermediate values.
+  func projectToNormalized(_ wp: Vector3) ->
```
dellaert, line 87: I don't think Camera should be doing the conversion to 3D camera coordinates. That is something Pose3 should do

dellaert (changes requested):
Cool! Some nits, but mainly I think you should move Pose transform to Pose3 struct?

Sources/SwiftFusion/Geometry/PinholeCamera.swift
tuwuhs: Do you mean to get rid of the pose transformation in the camera, so that `project()` takes a point in camera coordinate? If we do that then we won't be able to use the optimized custom derivative rule.

Sources/SwiftFusion/Geometry/Cal3_S2.swift
tuwuhs: Will the doc generator be able to parse inline documentation comments?
```swift
  public var fx: Double  /// Focal length in X direction.
  public var fy: Double  /// Focal length in Y direction.
  public var s: Double  /// Skew factor.
  public var u0: Double  /// Image center in X direction.
  public var v0: Double  /// Image center in Y direction.
```

Sources/SwiftFusion/Geometry/Cal3_S2.swift
tuwuhs: I try to make it more readable by splitting the u and v components into separate lines.
```swift
    Vector2(
      fx * np.x + s * np.y + u0,
      fy * np.y + v0)
```
vs.
```swift
    Vector2(fx * np.x + s * np.y + u0,
      fy * np.y + v0)
```
vs.
```swift
    Vector2(fx * np.x + s * np.y + u0, fy * np.y + v0)
```

Sources/SwiftFusion/Geometry/Cal3_S2.swift
tuwuhs: No, I haven't been able to successfully configure `swift-format`. It still complains about incompatible SwiftSyntax. I guess I need to dig deeper into that. Or perhaps can anyone point me in the right direction on how to set it up with s4tf toolchain in Linux?

Sources/SwiftFusion/Geometry/CameraCalibration.swift
tuwuhs: OK.

Sources/SwiftFusion/Geometry/Cal3_S2.swift
dellaert: IDK. Note I can't engage in long review conversations/back and forth. I don't have the time. Just re-request a review when you want me to look at it again, I will do one review now.

Sources/SwiftFusion/Geometry/Cal3_S2.swift
dellaert: Yes, my comment was to make it more readable by making a one-liner

Sources/SwiftFusion/Geometry/Cal3_S2.swift
dellaert: IDK

Sources/SwiftFusion/Geometry/PinholeCamera.swift
dellaert: No. It should *call* a method in Pose3 to do the transformation, which should be differentiable itself.

dellaert (changes requested):
Re-request when ready. Feel free to ignore comments with a reason (but then don't resolve) but avoid back and forth question/answer

dellaert:
Ping :-)

Sources/SwiftFusion/Geometry/PinholeCamera.swift
tuwuhs: I'm already using the `Pose3` group action operator `*` for doing the conversion to 3D camera coordinates. Since it's already differentiable, yes, it works as is without the custom derivative. But it's slow. Looking at the form of the Jacobian, there are repeated computations of 1/z and the normalized coordinates (x/z, y/z). By manually computing the Jacobian we can simplify the expression to reduce the number of redundant computations. I believe this is inline with what's done [in GTSAM](https://github.com/borglab/gtsam/blob/develop/doc/Mathematica/CalibratedCamera.nb).

tuwuhs:
Sorry for the delay, it took me a while to convince myself whether or not we really need that custom derivative.

Sources/SwiftFusion/Geometry/Cal3_S2.swift
tuwuhs: OK, done

dellaert (approved):
Cool! @marcrasi it's enlightening that Tuwuh found he needed to create a custom derivative for vjpProjectToNormalized, as just using differentiable is slow? Might be worth looking into why this is so.

Issue #194, opened 2020-10-15
Colab (Hosted Runtime) failed to compile SwiftFusion


ProfFan:
However, my local runtime using latest release (`RELEASE-0.11-rc2`) works.

SwiftFusion tag is `d9153d9ef93d25010a8089bedd70fa766424359f`.

```swift
swift-frontend: /swift-base/swift/lib/Sema/CSSimplify.cpp:9630: void swift::constraints::ConstraintSystem::recordPotentialHole(swift::TypeVariableType *): Assertion `typeVar' failed.
Stack dump:
[...]

PR #195, opened 2020-10-20
remove algorithms that are in penguin now

marcrasi:
https://github.com/saeta/penguin/pull/120 moved these into Penguin.

For some reason, defining them in both places causes a compiler crash.

This PR fixes https://github.com/borglab/SwiftFusion/issues/194.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ jobs:
         uses: actions/checkout@master
       - name: Install toolchain
         run: |
-          sudo xcode-select -s /Applications/Xcode_12_beta.app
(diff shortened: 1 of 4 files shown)


marcrasi:
Oh sorry, I'm gonna need to update the tests to import penguin now. I'll do this in an hour after a meeting,

marcrasi:
Ok it works now.

I also needed to change Xcode_beta to Xcode_12.2. Not sure exactly why. It might be related to some of the [recent announcements here](https://github.com/actions/virtual-environments/blob/main/images/macos/macos-10.15-Readme.md).

ProfFan (approved):
Thanks Marc!

PR #202, opened 2020-10-29
add RAE bee tracking model

marcrasi:
This adapts pieces of https://colab.sandbox.google.com/drive/1cE0gvS4KADvoX2xF6js6LopsE51vg9do#scrollTo=k_kr6IL2d-e0 into a module "BeeTracking" and a binary "BeeTrackingTool" that runs model training and factor graph inference.

My next step will be to try to profile it and make it faster. My initial attempts at profiling just say it's spending a bunch of time in "libx10", without giving any information about what it's doing there, and this isn't very useful. So I'll try to get some more useful profiles.

### DIFF
--- a/Examples/BeeTrackingTool/main.swift
+++ b/Examples/BeeTrackingTool/main.swift
@@ -1,0 +1,109 @@
+import ArgumentParser
+import BeeDataset
+import BeeTracking
+import PythonKit
+import SwiftFusion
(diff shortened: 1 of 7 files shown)


ProfFan:
@marcrasi  Feel free to merge

ProfFan (approved):

PR #203, opened 2020-10-29
generalize mechanism for custom linearizations

marcrasi:
This adds a `static func linearized` to the Factor protocol, with a default implementation that does the default linearization using AD. This allows conformances to customize the linearization. This replaces the ad-hoc code that we had in FactorsStorage.swift that gave certain specific factors custom linearizations.

I'm planning to use this for my performance work in https://github.com/borglab/SwiftFusion/pull/202, but it's complicated and self-contained enough that I thought it should be a separate PR too.

### DIFF
--- a/Sources/SwiftFusion/Inference/Factor.swift
+++ b/Sources/SwiftFusion/Inference/Factor.swift
@@ -45,6 +45,19 @@ public protocol VectorFactor: Factor {
   /// of `x`.
   func linearizableComponent(at x: Variables)
     -> (LinearizableComponent, LinearizableComponent.Variables)
+
+  /// Returns the linearizations of `factors` at `x`.
+  ///
+  /// The reason this method operates on a collection instead of a single element is so that the
+  /// result can be type-erased without paying a huge performance cost. The type-erasure lets the
+  /// implementation choose the most efficient type for storing the linearization, without
+  /// affecting the callers.
+  ///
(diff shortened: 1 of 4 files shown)


Sources/SwiftFusion/Inference/Factor.swift
```diff
@@ -246,3 +259,71 @@ where Head: Differentiable, Tail: DifferentiableVariableTuple {
...
+  /// affecting the callers.
+  public static func linearized<C: Collection>(_ factors: C, at x: VariableAssignments)
+    -> AnyGaussianFactorArrayBuffer where C.Element == Self
+  {
+    typealias TangentVector = LinearizableComponent.Variables.TangentVector
+    typealias Linearization<A> = Type<JacobianFactor<A, ErrorVector>>
+      where A: SourceInitializableCollection, A.Element: Vector & DifferentiableVariableTuple
+
+    // For small dimensions, we use a fixed size array in the linearization because the allocation
+    // and indirection overheads of a dynamically sized array are releatively big.
```
ProfFan, line 280: relatively

ProfFan (approved):
Great! Note we also need to migrate the ppca branch because I added two specializations there.

marcrasi:
> Great! Note we also need to migrate the ppca branch because I added two specializations there.
 
I merged master into that branch right after merging this, so they're combined now!

PR #212, opened 2020-11-11
OIST video tracks

marcrasi:
I uploaded 20 tracks to GCS. This PR adds code that downloads the tracks from GCS and that loads them into memory.

My process for creating the tracks was:
1. Build candidate tracks by joining the nearest bounding boxes in consecutive frames.
2. Manually go through the results and select valid subsequences of the candidate tracks.


ProfFan (approved):
Thanks Marc! I see one problem though, since we only have track X-Y-orientation here how can we do proper train-test separation?

ProfFan:
@marcrasi  Feel free to merge if you think this is ready :)

PR #214, opened 2020-11-12
Multivariate Gaussian

ProfFan:


Sources/SwiftFusion/Probability/MultivariateGaussian.swift
```diff
@@ -0,0 +1,71 @@
...
+    self.covariance_inv = pinv(regularized_cov)
+  }
+
+  /// Calculated the negative log likelihood of *one* data point
+  /// Note this is NOT normalized probability
+  @differentiable public func negativeLogLikelihood(_ sample: Tensor<Double>) -> Double {
+    precondition(sample.shape == self.dims)
+
+    let normalized = (sample - mean!).expandingShape(at: 1)
+    let t = matmul(normalized.transposed(), matmul(covariance_inv!, normalized))
```
dellaert, line 67: matmul has a transpose flag which is probably faster

dellaert (approved):
Great!

Sources/SwiftFusion/Probability/MultivariateGaussian.swift
ProfFan: done

PR #215, opened 2020-11-12
tracker evaluation metrics

marcrasi:
Functions and datastructures that take (dataset, tracker) and produce evaluation metrics:
* accuracy and robustness as in VOT 2020
* expected average overlap as in VOT 2020
* I'll add Pose/Orientation/Translation metrics in a later PR tomorrow.

You must install the "shapely" python library on your system to use this.

The metric calculators have tests and also I tried it on a real dataset and it all seems to work correctly.

Here is a notebook demonstrating usage: https://colab.sandbox.google.com/drive/1avFw27ZIU6fR1IZ4PAYHCqOCI5QX3_yV#scrollTo=QL9N6iijCZe_

### DIFF
--- a/Examples/BeeTrackingTool/main.swift
+++ b/Examples/BeeTrackingTool/main.swift
@@ -159,8 +159,6 @@ struct InferTrackRAE: ParsableCommand {
       return OrientedBoundingBox(
         center: prediction[poseID], rows: video.track[0].rows, cols: video.track[0].cols)
     }
-
-    printMetrics(groundTruth: videoSlice.track, prediction: boxes)
   }
 }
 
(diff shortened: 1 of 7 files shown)


ProfFan:
Great! BTW, I think we still need a way to sample the negative batches (background), could you make a function like `makeOISTBatch` but outputs background patches instead? Thanks in advance! @marcrasi

marcrasi:
Yes!

dellaert (approved):
Awesome!
Did not do a super-detailed review :-)
We need to prioritize getting some good results into the paper now, so reviewer can sign off on the publication approval.

PR #219, opened 2020-11-13
~6x tracking speedup by: removing redundant work, switching images to Float, and parallelizing patch

marcrasi:
With this PR, `swift run -c release OISTVisualizationTool naive-rae --box-id 8 --verbose` goes from ~6 minutes to ~1 minute for me.

Only like 30s of that is graph inference time.

Main changes:
* Store frames as Float, and do patch in Float, but do all other operations in Double (so that I don't have to change everything to Double, which would be a huge change).
* Store ArrayImage as the measurement in factors instead of Tensor so that we don't have to convert the Tensor to an ArrayImage every time we take a patch.
* Makes `patch` use multithreading.


ProfFan (approved):

dellaert:
6x times speed up, not bad!!!!

PR #220, opened 2020-11-13
Add PCA tracker

ProfFan:


Examples/OISTVisualizationTool/Util.swift
```diff
@@ -95,6 +95,58 @@ public func makeNaiveBayesRAETracker(
     })
 }
 
+/// Returns a tracking configuration for a tracker using PCA.
+///
+/// Parameter model: The PCA model to use.
+/// Parameter statistics: Normalization statistics for the frames.
+/// Parameter frames: The frames of the video where we want to run tracking.
+/// Parameter targetSize: The size of the target in the frames.
```
dellaert, line 103: copy/pasted from the wrong thing: fg and bg models missing

Examples/OISTVisualizationTool/main.swift
```diff
@@ -510,6 +506,153 @@ struct TrainRAE: ParsableCommand {
   }
 }
 
+/// PPCA with probabilistic model
+///
+/// Tracking with a Naive Bayes with RAE
+struct NaivePca: ParsableCommand {
```
dellaert, line 512: Hmm :-) I feel there is a lot of copy/paste here again, so I guess my previous request to split up in multiple functions might not have made it? By copy/pasting that mess we have created more work for us in the future.

dellaert (commented):
A bit grumpy but will relent to deadline pressure :-)

Examples/OISTVisualizationTool/main.swift
ProfFan: really don't have time, it's due to the inability to make type erased container for probability distributions

ProfFan: * in short time

PR #239, opened 2020-11-18
Feature/resample

dellaert:
Implemented low-variance resampling method from prob robotics book, see, conveniently https://robotics.stackexchange.com/questions/16093/why-does-the-low-variance-resampling-algorithm-for-particle-filters-work

There are a few other small changes to MCEM

### DIFF
--- a/Examples/OISTVisualizationTool/main.swift
+++ b/Examples/OISTVisualizationTool/main.swift
@@ -322,21 +322,11 @@ struct NaiveRae: ParsableCommand {
 
     if verbose { print("Fitting Naive Bayes model") }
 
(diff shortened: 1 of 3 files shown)


ProfFan (approved):

dellaert:
Test seems to fail because:
```
Fatal error: 'try!' expression unexpectedly raised an error: Python exception: No module named 'numpy': file PythonKit/Python.swift, line 673
```

ProfFan:
Oh I see. I think we need a `pip3 install --user numpy` in `.github/workflows` before running swift test.

Let's merge this and after that I can put up a PR that fixes this.

PR #242, opened 2020-11-23
clean up and update dependencies

marcrasi:
* github actions: stop using set-env because it got deprecated (https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/)
* swift-benchmark: use version instead of commit, so that we don't break when dependencies update their version of swift-benchmark
* swift-models: update to newer commit (can't use version because swift-models doesn't have version tags yet)
  * needed to update some calls to a newer signature
* swift-argument-parser: added dependency. we don't directly depend on it but we transitively depend on it, and for some reason swiftpm is getting confused unless I add it to the dependency list.


ProfFan (approved):
Thanks Marc :)

PR #243, opened 2020-11-25
Fixing CI by installing numpy etc

ProfFan:


ProfFan:
I'll merge to unblock :)

PR #245, opened 2020-11-26
[1/5] Fixed warnings and wrong PCAEncoder

ProfFan:
This is the fixes picked from the kitchen sink


dellaert (approved):
Great, but improve documentation of the IdentityProjection class, especially why it exists and what it does? Seems you just globally replaced Random with Identity at this point.

PR #247, opened 2020-11-26
Fix the EM importance sampling

ProfFan:
This PR should fix the EM importance sampling.

### DIFF
--- a/Sources/BeeTracking/TrackingLikelihoodModel.swift
+++ b/Sources/BeeTracking/TrackingLikelihoodModel.swift
@@ -60,7 +60,7 @@ public struct TrackingLikelihoodModel<Encoder: AppearanceModelEncoder, FG:Genera
   /// Calculate the negative log likelihood of the likelihood model of a patch
   @differentiable
   public func negativeLogLikelihood(of patch: Tensor<Double>) -> Double {
-    let encoded = encoder.encode(patch.expandingShape(at: 0)).squeezingShape(at: 0)
+    let encoded = encoder.encode(sample: patch)
     return (foregroundModel.negativeLogLikelihood(encoded) - backgroundModel.negativeLogLikelihood(encoded))
   }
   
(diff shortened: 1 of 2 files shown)


Sources/BeeTracking/TrackingLikelihoodModel.swift
```diff
@@ -56,6 +56,19 @@ public struct TrackingLikelihoodModel<Encoder: AppearanceModelEncoder, FG:Genera
               foregroundModel : FG(from: fgFeatures, given:p?.foregroundModel),
               backgroundModel : BG(from: bgFeatures, given:p?.backgroundModel))
   }
+  
+  /// Calculate the negative log likelihood of the likelihood model of a patch
+  @differentiable
+  public func negativeLogLikelihood(of patch: Tensor<Double>) -> Double {
+    let encoded = encoder.encode(patch.expandingShape(at: 0)).squeezingShape(at: 0)
```
dellaert, line 63: Works in batch?

Sources/BeeTracking/TrackingLikelihoodModel.swift
```diff
@@ -101,7 +114,16 @@ extension TrackingLikelihoodModel : McEmModel {
       // sample from noise model on manual pose
       var proposal = datum.obb.center
       proposal.perturbWith(stddev: Vector3(0.3, 8, 4.6))
-      return (1.0, proposal)
+      return (
+        // The weight of the proposal, which equals to the unnormalized probability of the likelihood:
+        self.unnormalizedProbability(
```
dellaert, line 119: I’d prefer to use a variable for the patch

dellaert (approved):

Sources/BeeTracking/TrackingLikelihoodModel.swift
ProfFan: Sure, I'll do that!

Sources/BeeTracking/TrackingLikelihoodModel.swift
ProfFan: @dellaert I am thinking about changing this interface from `encode()` to `encode(batch: )` and `encode(sample: )`, how is that?

ProfFan: done

PR #248, opened 2020-11-27
[WIP, 3/5] EM Tracking

ProfFan:
This is still WIP

### DIFF
--- a/Scripts/Fan10.swift
+++ b/Scripts/Fan10.swift
@@ -1,0 +1,88 @@
+import ArgumentParser
+
+import SwiftFusion
+import BeeDataset
+import BeeTracking
+import TensorFlow
+import PythonKit
+import Foundation
+
+typealias LikelihoodModel = TrackingLikelihoodModel<DenseRAE, MultivariateGaussian, GaussianNB>
+
+/// Fan10: RP Tracker, using the new tracking model
+struct Fan10: ParsableCommand {
+  @Option(help: "Run on track number x")
+  var trackId: Int = 0
+
+  @Option(help: "Run for number of frames")
+  var trackLength: Int = 80
+
+  @Option(help: "Size of feature space")
+  var featureSize: Int = 30
+
+  @Flag(help: "Training mode")
+  var training: Bool = false
+
+  func getTrainingDataEM(
+    from dataset: OISTBeeVideo,
+    numberForeground: Int = 3000,
+    numberBackground: Int = 3000
+  ) -> [LikelihoodModel.Datum] {
+    let bgBoxes = dataset.makeBackgroundBoundingBoxes(patchSize: (40, 70), batchSize: numberBackground).map {
+      (frame: $0.frame, type: LikelihoodModel.PatchType.bg, obb: $0.obb)
+    }
+    let fgBoxes = dataset.makeForegroundBoundingBoxes(patchSize: (40, 70), batchSize: numberForeground).map {
+      (frame: $0.frame, type: LikelihoodModel.PatchType.fg, obb: $0.obb)
+    }
+    
+    return fgBoxes + bgBoxes
+  }
+  
+  // Just runs an RP tracker and saves image to file
+  // Make sure you have a folder `Results/fan10` before running
+  func run() {
+    let kHiddenDimension = 100
+    let np = Python.import("numpy")
+    let dataDir = URL(fileURLWithPath: "./OIST_Data")
+
+    let (imageHeight, imageWidth, imageChannels) =
+      (40, 70, 1)
+    
+    var rae = DenseRAE(
+      imageHeight: imageHeight, imageWidth: imageWidth, imageChannels: imageChannels,
+      hiddenDimension: kHiddenDimension, latentDimension: featureSize
+    )
+
+    rae.load(weights: np.load("./oist_rae_weight_\(featureSize).npy", allow_pickle: true))
+
+    var generator = ARC4RandomNumberGenerator(seed: 42)
+    var em = MonteCarloEM<LikelihoodModel>(sourceOfEntropy: generator)
+    
+    let trainingDataset = OISTBeeVideo(directory: dataDir, length: 100)!
+    
+    let trainingData = getTrainingDataEM(from: trainingDataset)
+    
+    let trackingModel = em.run(with: trainingData, iterationCount: 3)
+    
+    let (fig, track, gt) = runProbabilisticTracker(
+      directory: dataDir,
+      encoder: rae,
+      onTrack: trackId, forFrames: trackLength, withSampling: true,
+      withFeatureSize: featureSize,
+      savePatchesIn: "Results/fan10"
+    )
+
+    /// Actual track v.s. ground truth track
+    fig.savefig("Results/fan10/fan10_track\(trackId)_\(featureSize).pdf", bbox_inches: "tight")
+    fig.savefig("Results/fan10/fan10_track\(trackId)_\(featureSize).png", bbox_inches: "tight")
+
+    let json = JSONEncoder()
+    json.outputFormatting = .prettyPrinted
+
+    let track_data = try! json.encode(track)
+    try! track_data.write(to: URL(fileURLWithPath: "Results/fan10/fan10_track_\(trackId)_\(featureSize).json"))
+
+    let gt_data = try! json.encode(gt)
+    try! gt_data.write(to: URL(fileURLWithPath: "Results/fan10/fan10_gt_\(trackId)_\(featureSize).json"))
+  }
+}
--- a/Scripts/Fan10.swift
+++ b/Scripts/Fan10.swift
@@ -55,18 +55,41 @@ struct Fan10: ParsableCommand {
 
     rae.load(weights: np.load("./oist_rae_weight_\(featureSize).npy", allow_pickle: true))
 
-    var generator = ARC4RandomNumberGenerator(seed: 42)
+    let generator = ARC4RandomNumberGenerator(seed: 42)
     var em = MonteCarloEM<LikelihoodModel>(sourceOfEntropy: generator)
     
     let trainingDataset = OISTBeeVideo(directory: dataDir, length: 100)!
     
     let trainingData = getTrainingDataEM(from: trainingDataset)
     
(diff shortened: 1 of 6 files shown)


Scripts/Fan10.swift
```diff
@@ -0,0 +1,111 @@
...
+    rae.load(weights: np.load("./oist_rae_weight_\(featureSize).npy", allow_pickle: true))
+
+    let generator = ARC4RandomNumberGenerator(seed: 42)
+    var em = MonteCarloEM<LikelihoodModel>(sourceOfEntropy: generator)
+    
+    let trainingDataset = OISTBeeVideo(directory: dataDir, length: 100)!
+    
+    let trainingData = getTrainingDataEM(from: trainingDataset)
+    
+    let trackingModel = em.run(
```
dellaert, line 65: Document! I think what you're trying to do here is to run EM but with a different scheme, where the RAE is not retrained? I think this a good hack for now, but now we have a weird mix between protocols and functions. I think we have to choose one or the other, and I like protocols *much* better. BUT, we have to think about how to allow for multiple EM schemes with the same underlying model.

Sources/BeeTracking/OISTBeeVideo+Batches.swift
```diff
@@ -62,6 +62,84 @@ extension OISTBeeVideo {
     return statistics.normalized(stacked)
   }
 
+  /// Returns a batch of locations of background
```
dellaert, line 65: I thought we already had this type of functionality. Where is the old code, and how is it different? No copy/pasta please!

Sources/BeeTracking/OISTBeeVideo+Batches.swift
```diff
@@ -62,6 +62,84 @@ extension OISTBeeVideo {
...
+        return (frame: frame, obb: OrientedBoundingBox(
+            center: location,
+            rows: patchSize.0, cols: patchSize.1))
+      }
+    }
+
+    return obbs
+  }
+  
+  /// Returns a batch of locations of background
```
dellaert, line 93: I thought we already had this type of functionality. Where is the old code, and how is it different? No copy/pasta please!

Sources/BeeTracking/OISTBeeVideo+Batches.swift
```diff
@@ -62,6 +62,84 @@ extension OISTBeeVideo {
...
+    /// Anything not completely overlapping labels
+    var deterministicEntropy = ARC4RandomNumberGenerator(seed: 42)
+    let frames = self.randomFrames(self.frames.count, using: &deterministicEntropy)
+
+    // We need `batchSize / frames.count` patches from each frame, plus the remainder of the
+    // integer division.
+    var patchesPerFrame = Array(repeating: batchSize / frames.count, count: frames.count)
+    patchesPerFrame[0] += batchSize % frames.count
+
+    let obbs = zip(patchesPerFrame, frames).flatMap { args -> [(frame: Tensor<Double>?, obb: OrientedBoundingBox)] in
```
dellaert, line 80: Document. What is this piece of code doing?

Sources/BeeTracking/OISTBeeVideo+Batches.swift
```diff
@@ -62,6 +62,84 @@ extension OISTBeeVideo {
...
+
+    var deterministicEntropy = ARC4RandomNumberGenerator(seed: 42)
+    let frames = self.randomFrames(self.frames.count, using: &deterministicEntropy)
+
+    // We need `batchSize / frames.count` patches from each frame, plus the remainder of the
+    // integer division.
+    var patchesPerFrame = Array(repeating: batchSize / frames.count, count: frames.count)
+    patchesPerFrame[0] += batchSize % frames.count
+
+    let obbs = zip(patchesPerFrame, frames).flatMap { args -> [(frame: Tensor<Double>?, obb: OrientedBoundingBox)] in
```
dellaert, line 110: document, what is this piece of code doing? This entire method is very long. Can you split it into simpler functions?

Sources/BeeTracking/OISTBeeVideo+Batches.swift
```diff
@@ -62,6 +62,84 @@ extension OISTBeeVideo {
...
+              continue
+            }
+          }
+
+          // The point was not rejected, so return it.
+          return location
+        }
+        fatalError("could not find backround location after \(attemptCount) attempts")
+      }
+      return locations.map { location -> (frame: Tensor<Double>?, obb: OrientedBoundingBox) in
```
dellaert, line 132: this piece of code is *almost* identical to lines 83-87. Make it a function

Sources/BeeTracking/ProbabilisticTracker.swift
```diff
@@ -5,6 +5,69 @@ import TensorFlow
 import PythonKit
 import Foundation
 
+/// Runs the random projections tracker
+/// Given a training set, it will train an RP tracker
+/// and run it on one track in the test set:
+///  - output: image with track and overlap metrics
+public func runProbabilisticTracker<
+  Encoder: AppearanceModelEncoder,
```
dellaert, line 13: Why not just template on likelihood model?

Sources/BeeTracking/ProbabilisticTracker.swift
```diff
@@ -5,6 +5,69 @@ import TensorFlow
...
+  likelihoodModel: TrackingLikelihoodModel<Encoder, ForegroundModel, BackgroundModel>,
+  onTrack trackIndex: Int,
+  forFrames: Int = 80,
+  withSampling samplingFlag: Bool = false,
+  withFeatureSize d: Int = 100,
+  savePatchesIn resultDirectory: String? = nil
+) -> (fig: PythonObject, track: [Pose2], groundTruth: [Pose2]) {
+  let testSetStart = 100
+  
+  // Create tracker
```
dellaert, line 27: make 27-36 into separate function

Sources/BeeTracking/ProbabilisticTracker.swift
```diff
@@ -5,6 +5,69 @@ import TensorFlow
...
+    foregroundModel: likelihoodModel.foregroundModel, backgroundModel: likelihoodModel.backgroundModel
+  )
+  
+  // Run the tracker and return track with ground truth
+  let (track, groundTruth) = createSingleTrack(
+    onTrack: trackIndex, withTracker: &tracker,
+    andTestData: testData, withSampling: samplingFlag
+  )
+  
+  // Now create trajectory and metrics plot
```
dellaert, line 48: make 48-67 into separate function

Sources/BeeTracking/ProbabilisticTracker.swift
```diff
@@ -113,13 +176,17 @@ public func trainProbabilisticTracker<Encoder: AppearanceModelEncoder>(
 /// Parameter statistics: Normalization statistics for the frames.
 /// Parameter frames: The frames of the video where we want to run tracking.
 /// Parameter targetSize: The size of the target in the frames.
-public func makeProbabilisticTracker<Encoder: AppearanceModelEncoder>(
+public func makeProbabilisticTracker<
+  Encoder: AppearanceModelEncoder,
```
dellaert, line 180: template on likelihood model

Sources/BeeTracking/TrackingLikelihoodModel.swift
```diff
@@ -57,6 +57,25 @@ public struct TrackingLikelihoodModel<Encoder: AppearanceModelEncoder, FG:Genera
...
   
+  /**
+   Train only the probabilistic models from a collection of foreground and background patches
+   - Parameters:
+    - encoder: the encoder object
+    - foregroundPatches: [...,H,W,C] batch of foreground patches to train with
+    - backgroundPatches: [...,H,W,C] batch of background patches to train with
+    - p: optional hyperparameters.
+   */
+  public init(with encoder: Encoder,
```
dellaert, line 68: see my comment above. I think a training a likelihood model with a given encoder is a different beast. Not a required change in this PR but we should discuss.

Sources/BeeTracking/TrackingLikelihoodModel.swift
```diff
@@ -76,14 +95,14 @@ extension TrackingLikelihoodModel : McEmModel {
   public enum PatchType { case fg, bg }
   
   /// As datum we have a (giant) image and a noisy manual label for an image patch
-  public typealias Datum = (frame: Tensor<Double>, type: PatchType, obb:OrientedBoundingBox)
+  public typealias Datum = (frame: Tensor<Double>?, type: PatchType, obb:OrientedBoundingBox)
```
dellaert, line 98: why?

Sources/SwiftFusion/Inference/MonteCarloEM.swift
```diff
@@ -93,10 +93,12 @@ public struct MonteCarloEM<ModelType: McEmModel> {
   
   /// Run Monte Carlo EM given unlabeled data
   public mutating func run(with data:[ModelType.Datum],
+                           modelInitializer: (_ data: [ModelType.Datum], _ entropySource: inout AnyRandomNumberGenerator) -> ModelType,
```
dellaert, line 96: As I said, I don't like this change at all. The inits were prescribed by the Protocol. This makes a weird in-between design which is neither here nor there.

dellaert (changes requested):
I might be changing my mind: I think we have to talk about switching back to a protocol-based EM and not check in this weird hybrid.

Sources/BeeTracking/ProbabilisticTracker.swift
ProfFan: The problem is likelihood model is a struct, so it cannot be a template argument. I think we need to talk on this.

ProfFan:
I feel that this is a mess. I'll close this and rework a better solution!

Sources/BeeTracking/TrackingLikelihoodModel.swift
ProfFan: hopefully saves memory

dellaert: Hmmm :-)
First, I don't see where you actually exploit this, or how this would work. Second, I though the whole point of value semantics and copy on write was that we did not have to care about this. Please revert or convince me beyond "hopefully".

PR #251, opened 2020-12-10
make everything compile on the stock toolchain

marcrasi:
This makes everything compile on the stock toolchain! Also everything still compiles on tensorflow-0.12 toolchain, so that you can continue working with that.

@ProfFan, let me know if it's okay for me to merge this with the tracking branch and also update the tracking branch to work with the stock toolchain. If it's okay, I'll do that next.

Note: The process for installing the stock toolchain with the required TensorFlow libraries is currently a bit complicated, so I wouldn't suggest trying it yourself yet. Soon we should have a simplified process. But if you really want to try it yourself, let me know and I can describe the process.

The required changes were:
* `import _Differentiation` to enable differentiation
* The stock toolchain seems a bit more strict about requiring `@differentiable` annotations on functions that are called by other `@differentiable` functions, so I added a few more `@differentiable` annotations.
* Conditionally import PythonKit only if it is available, because it's not available by default on stock toolchains.
* Delete the dependency on `swift-models` (because it doesn't compile on stock toolchains yet), and copy-paste the required code from `swift-models` into a `ModelSupport` module.


dabrahams:
Nice! I get: `warning: dependency 'swift-argument-parser' is not used by any target`.

Suggest you remove that from Package.swift

ProfFan (approved):
This is great!

marcrasi:
> Nice! I get: warning: dependency 'swift-argument-parser' is not used by any target.

> Suggest you remove that from Package.swift

I added a TODO. I can't do it now because there is some bug in SwiftPM so that it fails to build without it. It's fixed in very recent toolchains, so we should be able to remove it soon.

dabrahams:
How do I get a stock toolchain this works with?

PR #262, opened 2021-01-19
Removed the use of zeroTangentVector from PinholeCamera

acoadmarmon:
@dan-zheng I've manually implemented the zeroTangentVector functionality for the CameraCalibration object and Cal3_S2 which inherets it. I've run the suite of tests and it works well.

There is still one instance of zeroTangentVector that is being used in the codebase at https://github.com/borglab/SwiftFusion/blob/9eca609c10cdaeb19e34ca047363c183ddad7c3e/Sources/SwiftFusion/Core/LieGroup.swift#L170 that I've had some trouble replacing, but I'll work on it as well. All other uses in the codebase are not being used as Richard Wei mentioned in his issue.

Let me know your thoughts on the current implementation, as well as if you have suggestions on removing the LieGroup zeroTangentVector.

### DIFF
--- a/Sources/SwiftFusion/Geometry/Cal3_S2.swift
+++ b/Sources/SwiftFusion/Geometry/Cal3_S2.swift
@@ -39,6 +39,11 @@ extension Cal3_S2: CameraCalibration {
   }
 }
 
+extension Cal3_S2 {
+  public func zeroTangentVector() -> Cal3_S2.TangentVector {
+    return Vector5(0.0, 0.0, 0.0, 0.0, 0.0)
+  }
+}
 
 /// Manifold coordinate for Cal3_S2.
 public struct Cal3_S2Coordinate: Equatable {
--- a/Sources/SwiftFusion/Geometry/CameraCalibration.swift
+++ b/Sources/SwiftFusion/Geometry/CameraCalibration.swift
@@ -12,4 +12,6 @@ public protocol CameraCalibration: Differentiable {
   /// Converts from normalized coordinate to image coordinate.
   @differentiable
   func uncalibrate(_ np: Vector2) -> Vector2
+
+  func zeroTangentVector() -> TangentVector
 }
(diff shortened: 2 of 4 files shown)


dan-zheng (approved):
Thank you Andrew! Looks good to me if tests pass and correctness is unaffected.

dan-zheng:
I'll go ahead and merge, assuming that tests pass and correctness is unaffected.

dan-zheng:
> There is still one instance of zeroTangentVector that is being used in the codebase at
> 
> https://github.com/borglab/SwiftFusion/blob/9eca609c10cdaeb19e34ca047363c183ddad7c3e/Sources/SwiftFusion/Core/LieGroup.swift#L170
> 
> that I've had some trouble replacing, but I'll work on it as well. All other uses in the codebase are not being used as Richard Wei mentioned in his issue.

Thanks for mentioning this. Do you have ideas for fixing this remaining `zeroTangentVector` usage?

I think it should be possible by adding some conformance requirement on `LieGroupCoordinate.LocalCoordinate` (protocol requirement inherited from `ManifoldCoordinate.LocalCoordinate`) to a new protocol like `DifferentiableHasInstanceZeroTangentVector` (terrible name) defining `var zeroTangentVector: TangentVector`.

Pinging others in case they have time and see a better solution: @marcrasi @rxwei

acoadmarmon:
@dan-zheng I can work on this but it will take a few days given my current to-do list and the added complexity of changing zeroTangentVector for this object. Since this seems to be a blocker for another issue, if anyone has time to fix this in the shorter term I'd defer to them. I'll pick this back up on Sunday if it hasn't been solved by then.

dan-zheng:
Thanks for the quick update Andrew! Unless anyone (@rxwei) has higher urgency, I think fixing the remaining `zeroTangentVector` usage whenever you have time sounds great.

PR #264, opened 2021-02-10
Update RAE Parameters to achieve 1.0 Robustness

acoadmarmon:
@ProfFan review this PR to make sure all of the changes are valid.

### DIFF
--- a/Scripts/Andrew01.swift
+++ b/Scripts/Andrew01.swift
@@ -11,7 +11,6 @@ import PenguinStructures
 
 /// Andrew01: RAE Tracker
 struct Andrew01: ParsableCommand {
-  
   @Option(help: "Run for number of frames")
   var trackLength: Int = 80
   
--- a/Scripts/Andrew01.swift
+++ b/Scripts/Andrew01.swift
@@ -1,0 +1,116 @@
+import ArgumentParser
+
+import SwiftFusion
+import BeeDataset
+import BeeTracking
+import TensorFlow
+import PythonKit
+import Foundation
+
+import PenguinStructures
+
+/// Andrew01: RAE Tracker
+struct Andrew01: ParsableCommand {
+  
+  @Option(help: "Run for number of frames")
+  var trackLength: Int = 80
+  
+  @Option(help: "Size of feature space")
+  var featureSize: Int = 256
+
+  @Option(help: "Pretrained weights")
+  var weightsFile: String?
+
+  // Runs RAE tracker on n number of sequences and outputs relevant images and statistics
+  // Make sure you have a folder `Results/andrew01` before running
+  func run() {
+    let np = Python.import("numpy")
+    let kHiddenDimension = 512
+
+    let (imageHeight, imageWidth, imageChannels) =
+      (40, 70, 1)
+
+    
+    var rae = DenseRAE(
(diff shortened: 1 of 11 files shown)


Scripts/Andrew01.swift
```diff
@@ -11,14 +11,11 @@ import PenguinStructures
...
 struct Andrew01: ParsableCommand {
-  @Option(help: "Run on track number x")
-  var trackId: Int = 0
-  
   @Option(help: "Run for number of frames")
   var trackLength: Int = 80
   
   @Option(help: "Size of feature space")
-  var featureSize: Int = 5
+  var featureSize: Int = 256
```
ProfFan, line 18: This would make a 256x256 covariance matrix, what is the time cost compared to, for example, 16-d feature?

Sources/BeeTracking/Visualizations.swift
```diff
@@ -77,6 +77,50 @@ public func plotPatchWithGT(frame: Tensor<Float>, actual: Pose2, expected: Pose2
...
+
+/// plot Comparison image
+public func plotFrameWithPatches(frame: Tensor<Float>, actual: Pose2, expected: Pose2, firstGroundTruth: Pose2) -> (PythonObject, PythonObject) {
+  let plt = Python.import("matplotlib.pyplot")
+  let mpl = Python.import("matplotlib")
+
+  let (fig, ax) = plt.subplots(figsize: Python.tuple([8, 4])).tuple2
+  ax.imshow(frame.makeNumpyArray() / 255.0)
+  let actualBoundingBox = OrientedBoundingBox(center: actual, rows: 40, cols: 70)
+  ax.plot(Python.list(actualBoundingBox.corners.map{$0.x} + [actualBoundingBox.corners.first!.x]), Python.list(actualBoundingBox.corners.map{$0.y} + [actualBoundingBox.corners.first!.y]), "r-")
```
ProfFan: I think you don't need to explicitly convert to Python.list here, but anyway.

ProfFan (approved):
Great, thanks Andrew!

Scripts/Andrew01.swift
acoadmarmon: It is quite slow, but significantly increases model performance. I can run a quick time test if you'd like.

Sources/BeeTracking/Visualizations.swift
acoadmarmon: Ok I'll update this!

acoadmarmon: @ProfFan removed the call to Python.list

Scripts/Andrew01.swift
acoadmarmon: ~8.5 seconds to infer each frame for 256-d feature vs. ~8.2 seconds to infer each frame for 16-d feature. I guess this part is not the bottleneck?

ProfFan: This is very interesting. Maybe later we can do some profiling and see why is this the case.

Sources/BeeTracking/Visualizations.swift
ProfFan: Nice :)

